PR #131, opened 2024-05-15
Add Duplicate Account and Repair Connection UIs

bryceAebi:
Adds support for
1. In staging environments, breaking a connection
2. Unlinking connections
3. Displaying account status pills (connection broken / needs duplication confirmation)
4. Duplication confirmation/denial
5. Link repair flow

### DIFF
--- a/src/api/layer/linked_accounts.ts
+++ b/src/api/layer/linked_accounts.ts
@@ -6,9 +6,7 @@ export const getLinkedAccounts = get<
   {
     businessId: string
   }
->(
-  ({ businessId }) => `/v1/businesses/${businessId}/external-accounts`,
-)
+>(({ businessId }) => `/v1/businesses/${businessId}/external-accounts`)
 
 export const confirmConnection = post<
   Record<string, unknown>,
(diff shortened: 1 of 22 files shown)


src/api/layer/linked_accounts.ts
```diff
@@ -29,11 +36,12 @@ export const denyConnection = post<
     accountId: string
   }
 >(
-  ({businessId, accountId}) => `/v1/businesses/${businessId}/external-accounts/${accountId}/confirm`
+  ({ businessId, accountId }) =>
+    `/v1/businesses/${businessId}/external-accounts/${accountId}/deny`,
 )
 
 // TODO: not implemented yet in backend
```
doneel, line 43: We can drop this comment now, right?

src/components/LinkedAccounts/LinkedAccountsContent.tsx
```diff
@@ -35,22 +35,132 @@ export const LinkedAccountsContent = ({
...
+      {data?.map((account, index) => {
+        let pillConfig
+        if (account.requires_user_confirmation_as_of) {
+          pillConfig = {
+            text: 'Confirm account',
+            config: [
+              {
+                name: 'Oops... this is a duplicate account!',
+                action: async () => {
+                  // TODO: trigger some sort of loading spinner here
```
doneel, line 47: Looking for your opinion: is this something we should do right now? Can bring in design help if so.

doneel (approved):
Caveat: I haven't run this locally to inspect visually, but approving for now because looks straightforward!

src/api/layer/linked_accounts.ts
bryceAebi: This remains accurate. Right now we are directly using `unlinkPlaidItem`

PR #143, opened 2024-05-17
feat: add tasks list

ArkuVonSymfon:
https://github.com/Layer-Fi/layer-react/assets/32065451/8ae4846d-e260-4bbb-85b8-c8b0d934335f

Screenshot 2024-05-17 at 09 17 39

### DIFF
--- a/src/components/Tasks/Tasks.tsx
+++ b/src/components/Tasks/Tasks.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useContext } from 'react'
 import { TasksContext } from '../../contexts/TasksContext'
 import { useTasks } from '../../hooks/useTasks'
 import { Container } from '../Container'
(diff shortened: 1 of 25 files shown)


src/components/Tasks/Tasks.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { TasksPending } from '../TasksPending'
+
+export const Tasks = () => {
+  const contextData = useTasks()
+
+  return (
+    <TasksContext.Provider value={contextData}>
+      <Container name='tasks-component'>
+        <TasksHeader />
+        {contextData?.isLoading || !contextData.data ? (
```
tomaszantas, line 17: Not sure if you can do this that way... I think that ` {contextData?.isLoading ...}` is using data directly from `useTasks()` and not from `TasksContext.Provider`.

Let's split this into two components:

```ts
export const Tasks = () => {
  const contextData = useTasks()

  return (
    <TasksContext.Provider value={contextData}>
      <TasksComponent />
    </TasksContext.Provider>
  )
}

export const TasksComponent = () => {
  const contextData = useContext(TasksContext)

  return (
      <Container name='tasks-component'>
        <TasksHeader />
        {contextData?.isLoading || !contextData.data ? (
          <div className={`Layer__tasks__loader-container`}>
            <Loader />
          </div>
        ) : (
          <>
            {contextData.data.length > 0 && <TasksPending />}
            <TasksList />
          </>
        )}
      </Container>
  )
}
```

src/components/TasksListItem/TasksListItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+            />
+            <div className='Layer__tasks-list-item__actions'>
+              <Button
+                disabled={
+                  task.status === 'COMPLETED' || userResponse.length === 0
+                }
+                variant={ButtonVariant.secondary}
+                onClick={() => update(task.id, userResponse)}
+              >
+                Update
```
tomaszantas, line 85: We should use `Save` when no responses were sent so far to the task. And `Update` when response was already sent. Do we see from API if task contains response?

tomaszantas (changes requested):

tomaszantas (approved):
LGTM!

PR #150, opened 2024-06-03
Make descriptions & receipts togglable in bank txns

AndrewAcomb:


src/components/BankTransactions/BankTransactions.tsx
```diff
@@ -58,6 +60,8 @@ export const BankTransactions = ({
   asWidget = false,
   pageSize = 15,
   categorizedOnly = false,
+  showDescriptions = true,
+  showReceiptUploads = true,
```
doneel, line 64: Let's default these to false until we actually have this as a feature

doneel (approved):

PR #152, opened 2024-06-04
Drawer component

oliwia1003:
Drawer component.

Used in Accounting Overview View - will be removed from there.

### DIFF
--- a/src/components/Drawer/Drawer.tsx
+++ b/src/components/Drawer/Drawer.tsx
@@ -1,0 +1,17 @@
+import React from 'react';
+
+const DrawerBackground = ({ isOpen, onClose }: { isOpen: boolean; onClose: () => void }) => (
+  <div className={`Layer__drawer-background ${isOpen ? 'open' : ''}`} onClick={onClose}></div>
+);
+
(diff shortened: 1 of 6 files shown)


src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+const DrawerBackground = ({ isOpen, onClose }: { isOpen: boolean; onClose: () => void }) => (
+  <div className={`Layer__drawer-background ${isOpen ? 'open' : ''}`} onClick={onClose}></div>
+);
+
+export const Drawer = ({ isOpen, onClose, children }: { isOpen: boolean; onClose: () => void; children: React.ReactNode }) => {
+  return (
```
ArkuVonSymfon, line 8: You can use `classnames` here

`const drawerClassName = classNames(
      'Layer__drawer',
     isOpen && 'Layer__drawer--is-open'
    )`

And put it to div element `div className={drawerClassName}`

ArkuVonSymfon (approved):
Looks good but you can improve one thing

tomaszantas (approved):
LGTM!

PR #155, opened 2024-06-06
context for notifications

oliwia1003:
Toast state added to context layer.

### DIFF
--- a/src/providers/LayerProvider/LayerProvider.tsx
+++ b/src/providers/LayerProvider/LayerProvider.tsx
@@ -2,8 +2,7 @@ import React, {
   PropsWithChildren,
   useReducer,
   useEffect,
-  Reducer,
-  useState,
+  Reducer
 } from 'react'
 import { Layer } from '../../api/layer'
 import { ToastProps } from '../../components/Toast/Toast'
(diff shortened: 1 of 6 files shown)


src/providers/LayerProvider/LayerProvider.tsx
```diff
@@ -98,24 +114,28 @@ export const LayerProvider = ({
     usePlaidSandbox,
     onboardingStep: undefined,
     environment,
+    toasts: [],
   })
 
+  // const [toasts, setToasts] = useState<(ToastProps & { isExiting: boolean })[]>([]);
```
tomaszantas, line 120: ✂️

src/types/layer_context.ts
```diff
@@ -1,107 +1,125 @@
-import { Business, Category } from '../types'
-import { ExpiringOAuthResponse } from './authentication'
+import { ToastProps } from "../components/Toast/Toast";
```
tomaszantas, line 1: Possibly some eslint/prettier issue here... we don't use semicolons `;`

tomaszantas (commented):
Looks good in overall, but there is some prettier / eslint issue.
Take a look pls also on #156 <- I've added a way to dismiss toast on click. If all there looks good, feel free to merge (it's targeted to your branch, so after merging PR on Github, you may need to pull changes locally before making any other adjustments)

ArkuVonSymfon (commented):
Nothing more than prettier

PR #160, opened 2024-06-06
Add csv download buttons to bank transactions and pnl

AndrewAcomb:
fixes: LAY-721

Also removes the non-functional download button on the GL page

### DIFF
--- a/src/views/Reports/Reports.tsx
+++ b/src/views/Reports/Reports.tsx
@@ -1,6 +1,6 @@
-import React, { RefObject, useContext, useRef } from 'react'
+import React, { RefObject, useContext, useRef, useState } from 'react'
 import { Layer } from '../../api/layer'
-import { Button, ButtonVariant } from '../../components/Button'
+import { Button, ButtonVariant, RetryButton } from '../../components/Button'
 import { Container } from '../../components/Container'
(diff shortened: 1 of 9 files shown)


src/views/Reports/Reports.tsx
```diff
@@ -13,28 +17,57 @@ export interface ReportsPanelProps {
...
+              year: year,
+              month: month,
+            },
+          },
+        )
+        try {
+          const result = await getProfitAndLossCsv()
+          if (result?.data?.presignedUrl)
+            window.location.href = result.data.presignedUrl
+        } catch (e) {}
```
doneel, line 45: Nervous about errors - lots of surface area for silent failure.  We have an error notification pattern, can you try seeing if there's anything useful for this case?

doneel (approved):
Request to look into showing an error has happened but your judgmental call on what's worth it atm.

AndrewAcomb:
Screenshot 2024-06-06 at 6 29 18 PM

No longer silently failing

src/views/Reports/Reports.tsx
AndrewAcomb: Used the failed / retry button pattern from the bank transaction categorization approval button

PR #171, opened 2024-06-10
feat: support id-based categorization in categorization component

tomaszantas:
## Description

Use `accountId` in categorization by default and fallback to `stable_name`.


ArkuVonSymfon (approved):
LGTM!

PR #176, opened 2024-06-17
feat: control bank transactions list via hooks and props

tomaszantas:
## Description

1. Add optional `filters` prop to the `BankTransactions` so developers have greater control over the list of transactions.
2. Export `useBankTransactionsContext` so user can dynamically read and set data. 
3. Standarize using `useLayerContext` by moving it to `contexts/*`

## How this has been tested?

image

image


ArkuVonSymfon (approved):
LGTM!

PR #181, opened 2024-06-18
feat: drawer for tablets

tomaszantas:
## Description

1. Adjust sizing of drawer for "tablets"
2. improve drawer's closing animation
3. limit the width of mobile bank transactions list

## How this has been tested?

image

image

image

### DIFF
--- a/src/contexts/DrawerContext/DrawerContext.tsx
+++ b/src/contexts/DrawerContext/DrawerContext.tsx
@@ -6,4 +6,6 @@ export const DrawerContext = createContext<DrawerContextType>({
   content: undefined,
   setContent: () => undefined,
   close: () => undefined,
+  isClosing: false,
+  finishClosing: () => undefined,
 })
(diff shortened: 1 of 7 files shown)


src/contexts/DrawerContext/DrawerContext.tsx
```diff
@@ -6,4 +6,6 @@ export const DrawerContext = createContext<DrawerContextType>({
   content: undefined,
   setContent: () => undefined,
   close: () => undefined,
+  isClosing: false,
+  finishClosing: () => undefined,
```
ArkuVonSymfon, line 10: Isn't better to call this 'onDraweClose' or 'finishDrawerClose' or something like that? WDYT?

ArkuVonSymfon (approved):
Left on suggestion but we can go at it is

src/contexts/DrawerContext/DrawerContext.tsx
tomaszantas: I believe that it will become redundant to have `*Drawer*` in every action and context. And to avoid conflicts when using actions from other hooks, we can still do this:

```ts
const { close: closeDrawer } = useContext(DrawerContext)
```

PR #184, opened 2024-06-20
Add receipt and memo uploads

AndrewAcomb:


src/components/ExpandedBankTransactionRow/ExpandedBankTransactionRow.tsx
```diff
@@ -130,8 +131,12 @@ export const ExpandedBankTransactionRow = forwardRef<SaveHandle, Props>(
     const [height, setHeight] = useState<string | number>(0)
     const [isOver, setOver] = useState(false)
     const bodyRef = useRef<HTMLSpanElement>(null)
+    const [memoText, setMemoText] = useState('')
```
doneel, line 134: I would rather have this default to be `undefined` so that we can not set the memo if it's empty instead of empty string. but that should not block today's demo

doneel (approved):
all looks good so far!

doneel (approved):

PR #185, opened 2024-06-20
Daniel/2024 06 20 demo fixes

doneel:
1. Add account card height fixed 
image

2. `To Review` card loads, though the fix is hacky and not the right long term one
image

3. Remove monthly balance from Chart of accounts 
image


src/components/TransactionToReviewCard/TransactionToReviewCard.tsx
```diff
@@ -20,7 +20,17 @@ export const TransactionToReviewCard = ({
...
+    data,
+    isLoading,
+    error,
+    refetch,
+    activate: activateBankTransactions,
+  } = useBankTransactions()
+
+  useEffect(() => {
+    activateBankTransactions()
+  }, [])
```
AndrewAcomb, line 33: This doesn't work if you go to the Overview tab from another tab

AndrewAcomb (approved):
Not going to block but the spinner fix only conditionally works

PR #190, opened 2024-06-24
remove nonfunctional download buttons, rerender date picker

AndrewAcomb:
Fixes: LAY-771 , LAY-768


src/components/ProfitAndLossDatePicker/ProfitAndLossDatePicker.tsx
```diff
@@ -12,6 +12,7 @@ export const ProfitAndLossDatePicker = () => {
 
   return (
     <DateMonthPicker
+      key={dateRange.startDate.toISOString()}
```
AndrewAcomb, line 15: this makes the picker re-render when the month is changed

doneel (approved):

PR #200, opened 2024-06-25
fix: standarize sidebar headers

tomaszantas:
## Description

Standardize headings in sidebars.

## How this has been tested?

image

image

image

image

image

image

### DIFF
--- a/src/components/ProfitAndLossDetailedCharts/ProfitAndLossDetailedCharts.tsx
+++ b/src/components/ProfitAndLossDetailedCharts/ProfitAndLossDetailedCharts.tsx
@@ -66,9 +66,14 @@ export const ProfitAndLossDetailedCharts = ({
 
       <header className='Layer__profit-and-loss-detailed-charts__header--tablet'>
         <BackButton onClick={() => setSidebarScope(undefined)} />
-        <Text size={TextSize.lg} weight={TextWeight.bold} className='title'>
-          {humanizeTitle(theScope)}
-        </Text>
+        <div className='Layer__profit-and-loss-detailed-charts__head'>
+          <Text size={TextSize.lg} weight={TextWeight.bold} className='title'>
+            {humanizeTitle(theScope)}
+          </Text>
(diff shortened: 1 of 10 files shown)


src/components/ProfitAndLossDetailedCharts/ProfitAndLossDetailedCharts.tsx
```diff
@@ -66,6 +66,14 @@ export const ProfitAndLossDetailedCharts = ({
 
       <header className='Layer__profit-and-loss-detailed-charts__header--tablet'>
         <BackButton onClick={() => setSidebarScope(undefined)} />
+        <div className='Layer__profit-and-loss-detailed-charts__head'>
+          <Text size={TextSize.lg} weight={TextWeight.bold} className='title'>
```
ArkuVonSymfon, line 70: Should this className be just "title" or "date" ? Not like "Layer__title"?

src/styles/details_list.scss
```diff
@@ -3,6 +3,16 @@
   flex-direction: column;
 }
 
+.Layer__details-list__actions > .Layer__details-list__close-btn {
+  width: 36px;
+  height: 36px;
+  padding: 0;
+
+  @container (max-width: 1025px) {
```
ArkuVonSymfon, line 11: most small viewports have 1024 should we use 1025?

ArkuVonSymfon (approved):
Left some comments to double-check if that is correct

src/components/ProfitAndLossDetailedCharts/ProfitAndLossDetailedCharts.tsx
tomaszantas: I would go with plain `title` and `date`. `Layer__title` indicates on more globally used component.

src/styles/details_list.scss
tomaszantas: Fixed, thanks for catching this!

PR #206, opened 2024-06-28
Add cashflow summary rows

doneel:


src/components/StatementOfCashFlowRow/StatementOfCashFlowRow.tsx
```diff
@@ -125,6 +128,16 @@ export const StatementOfCashFlowRow = ({
           <span className='Layer__table-cell-content'>{amountString}</span>
         </td>
       </tr>
+      {summarize && (
```
doneel, line 131: This is to make sure we have a summary row *even if there aren't children*. i.e. when there are no investing activities, we should still show the total... even if that total is zero.

AndrewAcomb (approved):

PR #213, opened 2024-07-08
fix: Profit and Loss - chart colors - use similar colors for same categories

oliwia1003:
Zrzut ekranu 2024-07-8 o 10 52 24
Zrzut ekranu 2024-07-8 o 10 52 45
Zrzut ekranu 2024-07-8 o 10 53 06


tomaszantas (approved):
LGTM!

PR #222, opened 2024-07-16
Quickbooks integration

bryceAebi:
This PR includes:
1. Working OAuth linking/unlinking
2. Untested Quickbooks sync-from trigger

### DIFF
--- a/src/api/layer/quickbooks.ts
+++ b/src/api/layer/quickbooks.ts
@@ -1,0 +1,47 @@
+import { StatusOfSyncFromQuickbooks } from "../../types/quickbooks";
+import { get, post } from "./authenticated_http";
+
+export const syncFromQuickbooks = post<
+  Record<string, unknown>,
+  Record<string, unknown>,
+  { businessId: string }
+>(({ businessId }) => `/v1/businesses/${businessId}/quickbooks/sync-from`);
+
+export const statusOfSyncFromQuickbooks = get<
+  { data: StatusOfSyncFromQuickbooks },
+  { businessId: string }
+>(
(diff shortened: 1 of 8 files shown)


src/index.tsx
```diff
@@ -13,11 +13,14 @@ export { Onboarding } from './components/Onboarding'
...
 
+/* ------------------- Third Party Accounting Systems --------------- */
+export { Quickbooks } from './components/Quickbooks'
+
 /* ------------------ Reporting ------------------ */
 export { ProfitAndLoss } from './components/ProfitAndLoss'
 /* Profit and loss contains 6 sub-components:
  * - Chart
- * - Context
+ * - ContexCC
```
doneel, line 23: I think this was accidental

src/api/layer/quickbooks.ts
```diff
@@ -0,0 +1,47 @@
...
+  { data: StatusOfSyncFromQuickbooks },
+  { businessId: string }
+>(
+  ({ businessId }) => `/v1/businesses/${businessId}/quickbooks/sync-from/status`
+);
+
+export const initQuickbooksOAuth = post<
+  {
+    data: {
+      type: "Quickbooks_Authorization_Params";
```
doneel, line 20: I don't think you ned to include this in the body of the request

doneel (approved):
Looks good!

src/api/layer/quickbooks.ts
bryceAebi: The naming is a bit confusing, but this is actually the response body

PR #224, opened 2024-07-18
Daniel/lay 747 dedicated layout for moxie use case

doneel:


src/components/TasksHeader/TasksHeader.tsx
```diff
@@ -13,7 +14,7 @@ const ICONS = {
     badge: BadgeVariant.SUCCESS,
   },
   pending: {
-    icon: <ProgressIcon size={12} />,
+    icon: <AlertCircle size={12} />,
```
doneel, line 17: Preferred the non-spinner icon for the "work to be done" state. Spinner made me think it was stuck loading.

src/components/TasksHeader/TasksHeader.tsx
```diff
@@ -31,8 +32,7 @@ export const TasksHeader = ({
 }) => {
   const { data: tasks, refetch, error } = useContext(TasksContext)
 
-  const completedTasks = tasks?.filter(task => task.status === 'COMPLETED')
-    .length
+  const completedTasks = tasks?.filter(task => isComplete(task.status)).length
```
doneel, line 35: Many examples of this - there are multiple states we want to show as "completed" in the UI. Not intuitive, so not surprised you didn't have this info.

src/components/TasksList/TasksList.tsx
```diff
@@ -34,7 +35,7 @@ export const TasksList = ({ pageSize = 10 }: { pageSize?: number }) => {
       {sortedTasks && sortedTasks.length > 0 ? (
         <>
           {sortedTasks.map((task, index) => (
-            <TasksListItem key={index} task={task} index={index} />
+            <TasksListItem key={task.id} task={task} index={index} />
```
doneel, line 38: The ordering changes when we submit a task (which is good) but causes classic react bugs where the value of text in each box shifts when you submit.

src/components/TasksListItem/TasksListItem.tsx
```diff
@@ -16,20 +17,26 @@ export const TasksListItem = ({
   task: TaskTypes
   index: number
 }) => {
-  const [isOpen, setIsOpen] = useState(index === 0 ? true : false)
+  const [isOpen, setIsOpen] = useState(
+    index === 0 && !isComplete(task.status) ? true : false,
```
doneel, line 21: Don't want to default the first task to be open if it's already completed.

src/components/TasksListItem/TasksListItem.tsx
```diff
@@ -68,19 +75,22 @@ export const TasksListItem = ({
...
               onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) =>
                 setUserResponse(e.target.value)
               }
             />
             <div className='Layer__tasks-list-item__actions'>
               <Button
                 disabled={
-                  task.status === 'COMPLETED' || userResponse.length === 0
+                  userResponse.length === 0 ||
+                  userResponse === task.user_response
```
doneel, line 87: In addition to disabling the button for empty answers, this makes button disabled if the text hasn't changed from what we think the previous answer is. The effect is that when you open a completed task, the button is disabled until you change the text, then after submitting an update, the button goes back to disabled to help indicate that your answer was saved.

src/components/TasksListItem/TasksListItem.tsx
```diff
@@ -68,19 +75,22 @@ export const TasksListItem = ({
...
                 }
                 variant={ButtonVariant.secondary}
-                onClick={() => submitResponseToTask(task.id, userResponse)}
+                onClick={() => {
+                  submitResponseToTask(task.id, userResponse)
+                  setIsOpen(false)
```
doneel, line 92: After submitting, I want the task to close and go out of the way rather than stay open, which made it seem to me like I needed to do something else still.

PR #229, opened 2024-07-18
feat: make tasks collapsable

tomaszantas:
## Description

Make Tasks list collapsable - optional feature.

On Bookkeeping overview, make tasks collapsable only on "small" screens, when layout changes into single column.

## How this has been tested?

https://github.com/user-attachments/assets/bd2e1383-0b85-458c-b9a8-5dba6d0184d6


src/components/Tasks/Tasks.tsx
```diff
@@ -1,39 +1,73 @@
...
+  tasksHeader,
+  collapsable = false,
+  defaultCollapsed = false,
+}: {
+  tasksHeader?: string
+  collapsable?: boolean
+  defaultCollapsed?: boolean
+}) => {
   const { isLoading, data } = useContext(TasksContext)
+  const [open, setOpen] = useState(defaultCollapsed ? false : true)
```
doneel, line 44: I'd love to have the logic be: 

- If all tasks are complete: then default collapsed
- If there are any incomplete tasks, then default open
- This behavior can be overridden from the parent component

I _think_ we could solve this with this
```suggestion
  const [open, setOpen] = useState(defaultCollapsed ? data.tasks.filter(t => isComplete(t)).length = data.tasks.length)
```

doneel (approved):

src/components/Tasks/Tasks.tsx
tomaszantas: I had to add a bit more code, because `data` is empty at first painting. Change commited, feel free to merge PR

PR #231, opened 2024-07-23
feat: use inverted expenses on pnl chart

tomaszantas:
## Description

Show inverted expenses on pnl chart.

## How this has been tested?

image

### DIFF
--- a/src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
+++ b/src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
@@ -5,9 +5,7 @@ import {
   useProfitAndLossLTM,
 } from '../../hooks/useProfitAndLoss/useProfitAndLossLTM'
 import { centsToDollars } from '../../models/Money'
-import { ProfitAndLossSummary } from '../../types/profit_and_loss'
 import { isDateAllowedToBrowse } from '../../utils/business'
-import { capitalizeFirstLetter } from '../../utils/format'
 import { ProfitAndLoss as PNL } from '../ProfitAndLoss'
 import { Text } from '../Typography'
 import { Indicator } from './Indicator'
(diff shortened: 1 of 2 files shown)


src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
```diff
@@ -280,6 +282,19 @@ export const ProfitAndLossChart = () => {
     }
 
     return data
+      ?.map(x => {
+        if (x.operatingExpenses < 0 || x.uncategorizedOutflows < 0) {
```
doneel, line 286: In theory there's a case where `operatingExpenses < 0` but `uncategorizedOutflows > 0` and I think we would want to handle that separately, right?

src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
```diff
@@ -591,6 +606,38 @@ export const ProfitAndLossChart = () => {
           xAxisId='expenses'
           stackId='expenses'
         />
+        <Bar
```
doneel, line 609: After looking at the example image you posted in the PR description (very helpful), I think my preference would be to put the `uncategorizedOutflowsInverse` *above* the categorized expenses and categorized revenue. I think it feels better to have the profit dot visibly sit on top of the `categorized` bars and then have the uncategorized bars higher up.

doneel (approved):
A couple changes requested, but LGTM!

src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
tomaszantas: Yes, this handles such case. For example: `operatingExpenses < 0` but `uncategorizedOutflows > 0`

image

When both negative: `operatingExpenses < 0` but `uncategorizedOutflows < 0` (after change suggested in the next comment)

image

src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
tomaszantas: I've changed the order, screenshot attached to the previous comment

PR #232, opened 2024-07-23
fix: remove transaction after submit

tomaszantas:
## Description

Remove transaction from mobile list after successful submit.

## How this has been tested?


tomaszantas:
Changes implemented in #233

PR #233, opened 2024-07-23
feat: paginate bank transactions api calls

tomaszantas:
## Description

1. Use API pagination in Bank Transactions
2. Use `categorized` in API calls to Bank Transactions
3. using `categorized` in API allowed to simplify some logic so this PR containsa bunch of improvements
4. fix removing transactions after categorizing and improve animations

## Context

[LAY-830](https://linear.app/layerfi/issue/LAY-830/transactions-not-moved-to-categorized-on-mobile)
[LAY-842](https://linear.app/layerfi/issue/LAY-842/paginate-transaction-table)
[LAY-826](https://linear.app/layerfi/issue/LAY-826/approving-transactions-creates-thick-table-row-borders)

## How this has been tested?

https://github.com/user-attachments/assets/b6ee152a-9a82-46a4-bfd1-b957ce7a3338

https://github.com/user-attachments/assets/7d65483e-bd64-43d2-bf97-3e3e73ee36b1


src/styles/bank_transactions.scss
```diff
@@ -417,7 +418,7 @@
   }
 
   &.Layer__bank-transaction-row--removing {
-    display: none;
+    opacity: 0.99;
```
ArkuVonSymfon, line 421: Why are we using 0.99 opacity?

ArkuVonSymfon (approved):

src/styles/bank_transactions.scss
tomaszantas: Good point! we don't need this any longer. Previously, we used change on `opacity` or `top` to know when animation ends and remove permanently transaction from the list. Now, we use different approach.

PR #240, opened 2024-07-25
feat: datepicker drawer on mobile

ArkuVonSymfon:
## Context:

[LAY-847](https://linear.app/layerfi/issue/LAY-847/update-datepickers)
Screenshot 2024-07-26 at 15 54 18
Screenshot 2024-07-26 at 15 54 41
Screenshot 2024-07-26 at 15 54 49
Screenshot 2024-07-26 at 15 54 58


tomaszantas:
Looks good, two small suggestions:

image

Let's use shorter name in date picker for larger breakpoint, ie. Aug instead of August. (I see this works on smaller sizes)

Let's add more paddings on top and bottom

src/components/DatePicker/DatePicker.tsx
```diff
@@ -1,6 +1,9 @@
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import ReactDatePicker from 'react-datepicker'
 import 'react-datepicker/dist/react-datepicker.css'
+import { useSizeClass } from '../../hooks/useWindowSize'
```
tomaszantas, line 4: Try to use `useElementSize` which works on actual component instead of whole window. With `useElementSize` it will work better when we put few component in the same line.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -154,7 +195,31 @@ export const DatePicker = ({
...
+        onCalendarOpen={() => {
+          if (!isDesktop) {
+            document
+              .getElementById('Layer__datepicker__portal')
+              ?.classList.remove('Layer__datepicker__portal--closed')
+            document
+              .getElementById('Layer__datepicker__portal')
+              ?.classList.add('Layer__datepicker__portal--opened')
+          }
+          console.log('Calendar Opened')
```
tomaszantas, line 209: I think we can remove this

src/components/DatePicker/DatePicker.tsx
```diff
@@ -154,7 +195,31 @@ export const DatePicker = ({
...
+        onCalendarClose={() => {
+          if (!isDesktop) {
+            document
+              .getElementById('Layer__datepicker__portal')
+              ?.classList.add('Layer__datepicker__portal--closed')
+            document
+              .getElementById('Layer__datepicker__portal')
+              ?.classList.remove('Layer__datepicker__portal--opened')
+          }
+          console.log('Calendar Closed')
```
tomaszantas, line 220: Remove

tomaszantas:
One more thing - when used on actual mobile device, the native keyboard covers the drawer.

Here is how we can disable showing Keyboard and open picker programmatically:

```ts
const pickerRef = useRef<ReactDatePicker>(null)

<ReactDatePicker
        wrapperClassName={datePickerWrapperClassNames}
        ...
        onFocus={(e) => e.target.readOnly = true}
        onInputClick={() => {
          if (pickerRef.current) {
            pickerRef.current.setOpen(!pickerRef.current.isCalendarOpen())
          }
        }}
       ...
/>

```

src/components/DatePicker/DatePicker.tsx
ArkuVonSymfon: Hmm, I think it should works with window - we need to launch portal on mobile only so it should be not resize element (element has same width almost always). Will leave at it is or we can sync about that

tomaszantas (approved):
LGTM!

PR #242, opened 2024-07-28
Tasks navigation

doneel:
Three changes
1. Don't re-order tasks after completion: completed ones don't go to the bottom. This has nasty UX side effects where you can "complete" all your tasks and not realize there are multiple pages (it's not super obvious), so two additional changes...
2. Upon page open, we go to the first page with incomplete tasks.
3. Upon completing all the tasks on a page, automatically change to the next page.

### DIFF
--- a/src/components/TasksList/TasksList.tsx
+++ b/src/components/TasksList/TasksList.tsx
@@ -1,10 +1,19 @@
-import React, { useContext, useMemo, useState } from 'react'
+import React, { useContext, useEffect, useMemo, useState } from 'react'
 import { TasksContext } from '../../contexts/TasksContext'
 import SmileIcon from '../../icons/SmileIcon'
-import { isComplete } from '../../types/tasks'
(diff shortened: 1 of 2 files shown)


src/components/TasksList/TasksList.tsx
```diff
@@ -19,23 +28,51 @@ const TasksEmptyState = () => (
...
+  const goToNextPage = (task: TaskTypes) => {
+    const allComplete = sortedTasks
+      ?.filter(taskInList => taskInList.id !== task.id)
+      .every(task => isComplete(task.status))
+    if (allComplete) {
+      setCurrentPage(currentPage + 1)
+    }
+  }
+
+  console.log('TasksList', { tasks, sortedTasks, currentPage })
```
ArkuVonSymfon, line 63: [NIT] console log to remove

ArkuVonSymfon (approved):

PR #259, opened 2024-08-01
Use totalExpenses for pnl chart 

doneel:


src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
```diff
@@ -129,7 +129,13 @@ const getLoadingValue = (data?: ProfitAndLossSummaryData[]) => {
   return max === 0 ? 10000 : max * 0.6
 }
 
-export const ProfitAndLossChart = () => {
+export interface Props {
+  forceRerenderOnDataChange?: boolean
+}
+
+export const ProfitAndLossChart = ({
+  forceRerenderOnDataChange = false,
```
doneel, line 137: Turning the forced chart re-render off: this isn't meant to be on for clients in general as it has a nasty jittery effect. Just an option for debugging purposes.

PR #267, opened 2024-08-05
fix: total expenses

tomaszantas:
## Description

Follow-up to `totalExpenses` fix

## How this has been tested?

image


src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
```diff
@@ -244,9 +244,9 @@ export const ProfitAndLossChart = ({
     name: getMonthName(pnl),
     revenue: pnl?.income || 0,
     revenueUncategorized: pnl?.uncategorizedInflows || 0,
-    expenses: -(pnl?.operatingExpenses || 0),
```
tomaszantas, line 247: ^ this is the crucial part.. other changes is just about renaming variables

doneel (approved):
thank you I was too cowardly to make this renaming change but this is better

PR #272, opened 2024-08-06
feat: change profit and loss table

ArkuVonSymfon:
https://github.com/user-attachments/assets/f045443e-be59-40c7-9c06-8564ceb4f6a4

### DIFF
--- a/src/components/ProfitAndLossTable/ProfitAndLossTable.tsx
+++ b/src/components/ProfitAndLossTable/ProfitAndLossTable.tsx
@@ -1,8 +1,12 @@
-import React, { useContext } from 'react'
-import { Direction } from '../../types'
+import React, { useContext, useEffect } from 'react'
+import { SidebarScope } from '../../hooks/useProfitAndLoss/useProfitAndLoss'
+import { useTableExpandRow } from '../../hooks/useTableExpandRow'
+import PieChart from '../../icons/PieChart'
+import { BalanceSheet, Direction, LineItem } from '../../types'
 import { Loader } from '../Loader'
 import { ProfitAndLoss } from '../ProfitAndLoss'
 import { ProfitAndLossRow } from '../ProfitAndLossRow'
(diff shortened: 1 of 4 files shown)


src/components/ProfitAndLossTable/ProfitAndLossTable.tsx
```diff
@@ -47,95 +51,139 @@ export const ProfitAndLossTable = ({
...
+    rowIndex: number,
+    scope?: SidebarScope,
+    setSidebarScope?: (view: SidebarScope) => void,
+    variant?: 'default' | 'summation',
+  ): React.ReactNode => {
+    const expandable = !!lineItem.line_items && lineItem.line_items.length > 0
+
+    const expanded = expandable ? isOpen(rowKey) : true
+
+    const showChildren = expanded
```
tomaszantas, line 67: `showChildren` is not necessary, we can just use `expanded` in the code further

src/components/ProfitAndLossTable/ProfitAndLossTable.tsx
```diff
@@ -18,16 +21,17 @@ type Props = {
...
   const {
     data: actualData,
     isLoading,
     setSidebarScope,
   } = useContext(ProfitAndLoss.Context)
+  const { isOpen, setIsOpen } = useTableExpandRow()
+
+  useEffect(() => {
+    setIsOpen(['income'])
+  }, [])
```
tomaszantas, line 34: Forcing another repainting and updating is not optimal. Let's change `useTableExpandRow` so it can take prop like `defaultOpen` and initiate the state right away.

tomaszantas: We want to expand all first-level rows, so we need to use `['income', 'cost_of_goods_sold', 'expenses']`

tomaszantas (approved):
LGTM!

PR #275, opened 2024-08-06
fix: toggle thumb not updated when options disabled

tomaszantas:
## Description

Toggle thumb size and position was not updated when all options are disabled.

image

## How this has been tested?

image


doneel:
Thank you for catching!

doneel (approved):

PR #276, opened 2024-08-06
Handle null categorization_flow + account institution

AndrewAcomb:
Screenshot 2024-08-06 at 12 16 32 PM
Screenshot 2024-08-06 at 12 16 24 PM

Was getting an unhandled error for a business with transactions with null categorization flows and realized that this was not accounted for on the front-end.

Also included in this PR: falling back to the linked bank account name when the institution is null

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@layerfi/components",
-  "version": "0.1.43",
+  "version": "0.1.45",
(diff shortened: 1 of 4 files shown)


doneel (approved):
We still don't have balances for these accounts right? I think we'd have to manually enter those.

PR #290, opened 2024-08-13
feat: adjust usage of uncategorized in profit and loss

tomaszantas:
## Description

1. Show uncategorized values in summary cards
2. Show uncategorized in the tooltip on Profit and Loss chart
3. don't sum up uncategorized to "Total" in detailed charts

### DIFF
--- a/src/utils/profitAndLossUtils.ts
+++ b/src/utils/profitAndLossUtils.ts
@@ -63,10 +63,18 @@ export const humanizeTitle = (sidebarView: SidebarScope) => {
 export const applyShare = (
   items: LineBaseItem[],
   total: number,
+  exclude?: string[],
 ): LineBaseItem[] => {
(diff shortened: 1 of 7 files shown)


src/utils/profitAndLossUtils.ts
```diff
@@ -4,13 +4,15 @@ import { ProfitAndLoss } from '../types/profit_and_loss'
...
-    item.value < 0
+    (
+      item.is_contra ||
+      item.value === undefined ||
+      item.value === null ||
+      isNaN(item.value) ||
+      item.value === -Infinity ||
+      item.value === Infinity
+    )
+    // || item.value < 0
```
ArkuVonSymfon, line 15: Should we keep that?

ArkuVonSymfon (approved):

PR #291, opened 2024-08-14
fix: add profil and loss provider to export

ArkuVonSymfon:
## Log

Previously, it was necessary to have a `TableProvider` before using the profit and loss table. This requirement has now been removed. Instead, the `ProfitAndLoss.Table` is directly exported and can be used without needing a separate `TableProvider`.


tomaszantas (approved):

src/components/ProfitAndLossTable/index.tsx
```diff
@@ -1 +1,2 @@
-export { ProfitAndLossTable } from './ProfitAndLossTable'
```
doneel, line 1: Appreciate the backwards compatible export

doneel (approved):
Makes sense

PR #303, opened 2024-08-16
Andrew/swr key all params

AndrewAcomb:


src/hooks/useBankTransactions/useBankTransactions.tsx
```diff
@@ -304,16 +324,13 @@ export const useBankTransactions: UseBankTransactions = params => {
   }
 
   const removeAfterCategorize = (bankTransaction: BankTransaction) => {
-    /*
-    Removed 2024-08-12 by @doneel as part of unifying transaction lists in LAY-298
     const updatedData = rawResponseData?.map(page => {
```
doneel, line 327: This block is still going to cause us problems: we only want to do the client-side removal animation from the list a filter no longer applies. The only situation I can think of is if 
1. You have a filter set to uncategorized only.
2. You categorize a transaction.

None of the other data on a transaction that can be used by filters (`amount`, `account`, `direction`, `dateRange`) can be affected by user action.


So I think we'll have to do a special one-off case here where we only want to apply this affect in that situation. In my ideal world we have generalized method of "did filters previously _allow_ this transaction but now should _disallow_", but that would require
1. Keeping our client-side filter logic
2. Predicting the state of a transaction before we get a response back from the server.

This is to say, I don't think we're going to get a super generalizable way of doing this, so we should basically hardcode 
```ts
    if (
      filters?.categorizationStatus === DisplayState.review && 
      ReviewCategories.includes(bankTransaction.categorization_status)
    ) {
      //do the removal
    }
```

doneel (commented):
Only found one issue (always removing from the list when categorizing) - looking great so far.

src/hooks/useBankTransactions/useBankTransactions.tsx
AndrewAcomb: Turns out I needed to do this multiple places so I added it as shouldHideAfterCategorize in useBankTransactions. I used it and now we should only do the client-side removal animation when appropriate.

src/hooks/useBankTransactions/useBankTransactions.tsx
```diff
@@ -57,18 +51,52 @@ export const useBankTransactions: UseBankTransactions = params => {
       return [false, undefined]
     }
 
-    if (!prevData?.meta?.pagination?.cursor) {
+    if (_index === 0) {
```
tomaszantas, line 54: NIT:  Underscore can be removed from `_index` here:
 
```suggestion
    if (index === 0) {
```

... and in line 49:

```ts
const getKey = (index: number, prevData: any) => {
```

src/hooks/useBankTransactions/useBankTransactions.tsx
```diff
@@ -57,18 +51,52 @@ export const useBankTransactions: UseBankTransactions = params => {
...
-        businessId && auth?.access_token && `bank-transactions-${businessId}`,
+        businessId &&
+          auth?.access_token &&
+          `bank-transactions${
+            filters?.categorizationStatus
+              ? `-categorizationStatus-${filters.categorizationStatus}`
+              : `-categorizationStatus-${DisplayState.all}`
+          }${
+            filters?.direction?.length === 1
+              ? `-startDate-${filters.direction[0]}`
```
tomaszantas, line 64:
```suggestion
              ? `-direction-${filters.direction.join('-'}`
```

src/hooks/useBankTransactions/useBankTransactions.tsx
```diff
@@ -57,18 +51,52 @@ export const useBankTransactions: UseBankTransactions = params => {
...
         auth?.access_token &&
-        `bank-transactions-${businessId}-${prevData.meta.pagination.cursor}`,
-      prevData.meta.pagination.cursor,
+        `bank-transactions${
+          filters?.categorizationStatus
+            ? `-categorizationStatus-${filters.categorizationStatus}`
+            : `-categorizationStatus-${DisplayState.all}`
+        }${
+          filters?.direction?.length === 1
+            ? `-startDate-${filters.direction[0]}`
```
tomaszantas, line 88:
```suggestion
            ? `-direction-${filters.direction.join('-'}`
```

src/hooks/useBankTransactions/useBankTransactions.tsx
```diff
@@ -326,18 +363,39 @@ export const useBankTransactions: UseBankTransactions = params => {
...
 
+  const getCacheKey = (txnFilters?: BankTransactionFilters) => {
+    console.log(txnFilters)
+    return `bank-transactions${
+      txnFilters?.categorizationStatus
+        ? `-categorizationStatus-${txnFilters.categorizationStatus}`
+        : `-categorizationStatus-${DisplayState.all}`
+    }${
+      txnFilters?.direction?.length === 1
+        ? `-startDate-${txnFilters.direction[0]}`
```
tomaszantas, line 374:
```suggestion
        ? `-direction-${filters.direction.join('-'}`
```

doneel (approved):

PR #304, opened 2024-08-16
Daniel/move reset hook to layer context

doneel:


src/hooks/useDataSync/useDataSync.tsx
```diff
@@ -85,16 +85,16 @@ export const useDataSync: UseDataSync = () => {
...
+    const now = Date.now()
     setSyncTimestamps({
-      [DataModel.BALANCE_SHEET]: initialTimestamp,
-      [DataModel.CHART_OF_ACCOUNTS]: initialTimestamp,
-      [DataModel.JOURNAL]: initialTimestamp,
-      [DataModel.LEDGER_ACCOUNTS]: initialTimestamp,
-      [DataModel.LINKED_ACCOUNTS]: initialTimestamp,
-      [DataModel.PROFIT_AND_LOSS]: initialTimestamp,
-      [DataModel.STATEMENT_OF_CASH_FLOWS]: initialTimestamp,
+      [DataModel.BALANCE_SHEET]: now,
```
doneel, line 90: Setting these to the `initialTimestamp` makes it an _earlier_ time than now - when what we really want is to think that the data has been synced more _recently_.

src/hooks/useDataSync/useDataSync.tsx
```diff
@@ -85,16 +85,16 @@ export const useDataSync: UseDataSync = () => {
...
+      [DataModel.CHART_OF_ACCOUNTS]: now,
+      [DataModel.JOURNAL]: now,
+      [DataModel.LEDGER_ACCOUNTS]: now,
+      [DataModel.LINKED_ACCOUNTS]: now,
+      [DataModel.PROFIT_AND_LOSS]: now,
+      [DataModel.STATEMENT_OF_CASH_FLOWS]: now,
     })
-    setReadTimestamps({})
```
doneel, line 97: If we reset these time `undefined`, then the components think we don't have stale data _because we don't have any data_. 

We want to communicate that this data has been updated like any other data update, so we leave the read timestamps in place so the cache thinks there _is_ data, it's just out of date and needs to be refreshed.

src/providers/LayerProvider/LayerProvider.tsx
```diff
@@ -360,6 +366,7 @@ export const LayerProvider = ({
           read,
           syncTimestamps,
           readTimestamps,
+          expireDataCaches: resetCaches,
```
doneel, line 369: Renamed this `expiration` to more tightly correlate with what this does - leaves the caches in place just marks them out of date.

doneel, line 369: This looks like a trivial change, but it's actually important to export this _in the `<LayerProvider>`_ as opposed to a standalone hook, otherwise we'll get two separate instances of the timestamps cache!

AndrewAcomb (approved):

PR #308, opened 2024-08-19
dont call match endpoint for confirming existing matched transfers

AndrewAcomb:
Fixes: LAY-947 (partially)- backend changes needed to make categorization & matching idempotent

### DIFF
--- a/src/hooks/useBankTransactions/useBankTransactions.tsx
+++ b/src/hooks/useBankTransactions/useBankTransactions.tsx
@@ -347,10 +347,7 @@ export const useBankTransactions: UseBankTransactions = params => {
   const shouldHideAfterCategorize = (
     bankTransaction: BankTransaction,
   ): boolean => {
-    return (
-      filters?.categorizationStatus === DisplayState.review &&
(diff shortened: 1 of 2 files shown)


src/hooks/useBankTransactions/useBankTransactions.tsx
```diff
@@ -347,10 +347,7 @@ export const useBankTransactions: UseBankTransactions = params => {
   const shouldHideAfterCategorize = (
     bankTransaction: BankTransaction,
   ): boolean => {
-    return (
-      filters?.categorizationStatus === DisplayState.review &&
-      ReviewCategories.includes(bankTransaction.categorization_status)
-    )
```
doneel, line 353: This seems probably fine, but just checking why this is? Is it hiding matches?

doneel (approved):

src/hooks/useBankTransactions/useBankTransactions.tsx
AndrewAcomb: This was being called *after* a transaction was categorized/matched, so the `ReviewCategories.includes(bankTransaction.categorization_status)` was never true

PR #311, opened 2024-08-20
fix: reset bank transactions page on filters change

tomaszantas:
## Description

Bank transactions: Set page number to `1` when filters change.

## How this has been tested?

https://github.com/user-attachments/assets/cf345b32-f7d0-4844-b299-fcb153117a7d


doneel (approved):
This makes perfect sense to me, so simple.

PR #312, opened 2024-08-20
feat: profit and loss chart - sync and empty states

tomaszantas:
## Description

Support syncing and no-data states on profit and loss chart.

## How this has been tested?

1. NO syncing and NO data

_Bars are just gray placeholders with no action on click_

image

2. NO syncing and data

image

3. Syncing and NO data

_Bars have same animation as in loading state_

image

4. Syncing and data

image

### DIFF
--- a/src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
+++ b/src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
@@ -1,5 +1,6 @@
 import React, { useContext, useEffect, useMemo, useState } from 'react'
 import { useLayerContext } from '../../contexts/LayerContext'
+import { useLinkedAccounts } from '../../hooks/useLinkedAccounts'
 import {
   ProfitAndLossSummaryData,
   useProfitAndLossLTM,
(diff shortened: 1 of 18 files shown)


src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
```diff
@@ -174,6 +176,28 @@ export const ProfitAndLossChart = ({
...
+          x.income !== 0 ||
+          x.costOfGoodsSold !== 0 ||
+          x.grossProfit !== 0 ||
+          x.operatingExpenses !== 0 ||
+          x.profitBeforeTaxes !== 0 ||
+          x.taxes !== 0 ||
+          x.totalExpenses !== 0,
+      ),
+    )
+  }, [data])
```
tomaszantas, line 192: Not sure how to capture that there is no data... Above, I'm looking for any non-zero value

doneel (changes requested):
@tomaszantas I'm seeing the absolute positioning behave rudely on different layouts. Here's an example of a client page layout where the pnl chart is on a different page layout than our default page. It shows up over a different component (and gets cut off, but I circled it)
image

tomaszantas:
@doneel I fixed this in above [commit](https://github.com/Layer-Fi/layer-react/pull/312/commits/97709e02dac0ba63c0497ce60c8dc6f5039d03f0) by adding `position: relative` to the parent component.

doneel (approved):

PR #313, opened 2024-08-20
fix: allow for cents in journal

tomaszantas:
## Description

Allow for cents in Journal form. It doesn't allow for negative values and use only 2 decimal places.

## How this has been tested?

https://github.com/user-attachments/assets/a7f43e3b-f64d-4556-aa8e-78878e387e43


doneel (approved):
Working excellent!

PR #314, opened 2024-08-20
move SWR cache from page global to components-global

doneel:


src/providers/LayerProvider/LayerProvider.tsx
```diff
@@ -237,7 +235,6 @@ export const LayerProvider = ({
       params: { businessId },
     }),
     {
-      ...defaultSWRConfig,
```
doneel, line 240: Removing the explicit setting of the default SWR config so it uses the single global config we define in the `SWRConfig`.

src/providers/LayerProvider/LayerProvider.tsx
```diff
@@ -348,7 +345,7 @@ export const LayerProvider = ({
   const drawerContextData = useDrawer()
 
   return (
-    <SWRConfig value={defaultSWRConfig}>
+    <SWRConfig value={{ ...defaultSWRConfig, provider: () => new Map() }}>
```
doneel, line 348: Every time we re-initialize this `<SWRConfig`, we'll reset the cache. That means if you unmount and remount the component, you don't need to actually call `expireCaches`, we'll just automatically refresh everything.

src/types/layer_context.ts
```diff
@@ -33,7 +33,7 @@ export type LayerContextHelpers = {
   touch: (model: DataModel) => void
   read: (model: DataModel, cacheKey: string) => void
   syncTimestamps: Partial<Record<DataModel, number>>
-  readTimestamps: Partial<Record<DataModel, number>>
+  readTimestamps: Partial<Record<string, { t: number; m: DataModel }>>
```
doneel, line 36: This was previously wrong, just noticed it today while debugging.

AndrewAcomb (approved):

AndrewAcomb (approved):
Works well for me

PR #315, opened 2024-08-20
add onTransactionsFetched callback

AndrewAcomb:
For our purposes I think it is preferable for this to be called too often than not often enough. Thus I am just calling it every time we poll.

### DIFF
--- a/src/hooks/useBankTransactions/useBankTransactions.tsx
+++ b/src/hooks/useBankTransactions/useBankTransactions.tsx
@@ -414,7 +414,6 @@ export const useBankTransactions: UseBankTransactions = params => {
     if (refreshTrigger !== -1) {
       refetch()
       refetchAccounts()
-      eventCallbacks?.onTransactionsFetched?.()
     }
   }, [refreshTrigger])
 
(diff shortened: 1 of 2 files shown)


src/hooks/useBankTransactions/useBankTransactions.tsx
```diff
@@ -414,6 +414,7 @@ export const useBankTransactions: UseBankTransactions = params => {
     if (refreshTrigger !== -1) {
       refetch()
       refetchAccounts()
+      eventCallbacks?.onTransactionsFetched?.()
```
doneel, line 417: I am on board with the principle of erring on the side of over communicating, but this feels a LITTLE extreme since we do this with polling a lot. 

I'd propose we at least move this down to line ~441 where we only would trigger the callback if the data actually changes.

doneel (approved):
Approving with reservations.

src/hooks/useBankTransactions/useBankTransactions.tsx
AndrewAcomb: Makes sense- done.

PR #318, opened 2024-08-22
remove both txns when confirming transfer

AndrewAcomb:
Fixes: LAY-824


doneel (approved):
Looks good to me. @tomaszantas this may take care of the transfer matching work.

PR #322, opened 2024-08-27
Tom/pnl comparison api

tomaszantas:


src/components/ProfitAndLossCompareOptions/ProfitAndLossCompareOptions.tsx
```diff
@@ -71,11 +71,12 @@ export const ProfitAndLossCompareOptions = ({
   const tagComparisonSelectOptions = tagComparisonOptions.map(
     tagComparisonOption => {
       return {
-        value: tagComparisonOption.tagFilters,
+        value: JSON.stringify(tagComparisonOption.tagFilters),
```
tomaszantas, line 74: `tagFilters` is an object so ReactSelect doesn't work well with matching `defaultTagFilter` to the one of `tagComparisonOptions`, and it recognizes these two as different options.
The solution is to stringify `tagFilters`

doneel: Ah brilliant. Thank you

doneel (approved):

PR #335, opened 2024-09-05
feat: add descriptions into bank transactions categories

ArkuVonSymfon:
https://github.com/user-attachments/assets/4c54a766-a4b4-43ce-bb14-e81a7fa0504a


src/components/ActionableList/ActionableList.tsx
```diff
@@ -33,9 +34,20 @@ export const ActionableList = <T,>({
           className={classNames(
             x.secondary && 'Layer__actionable-list-item--secondary',
             x.asLink && 'Layer__actionable-list-item--as-link',
+            selectedId === x.id && 'Layer__actionable-list__item--selected',
           )}
         >
-          <Text>{x.label}</Text>
+          <div className='Layer__actionable-list__content'>
+            <Text size={TextSize.sm}>{x.label}</Text>
+            {x.description && x.label !== 'See all categories' && (
```
tomaszantas, line 42: I would rather go with something more generic instead of checking specific text: `"See all categories"`. Maybe we can add additional prop to the option?

AndrewAcomb (approved):
Will want to do some more cleanup later, but this is needed for a customer.

PR #346, opened 2024-09-15
Daniel/hide reversed ledger entries

doneel:
And also use journal entry numbers whenever possible.
image

One thing that's not perfect is that there's two situations where we don't support journal entry numbers
1. Old journal entries before ~2 months ago when we rolled out the entry numbers.
2. There are codepaths (e.g. payouts) which don't go through the journal entry number codepath. I would like to refactor it to be included in the future.


AndrewAcomb (approved):

tomaszantas (approved):
LGTM!

PR #347, opened 2024-09-16
flag for making template accounts uneditable

doneel:
When setting is enabled, it adds an uneditable state (with a tooltip) for accounts with a stable name, i.e. accounts created by  the system.
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@layerfi/components",
-  "version": "0.1.61",
+  "version": "0.1.62",
   "description": "Layer React Components",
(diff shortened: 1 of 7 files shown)


AndrewAcomb (approved):

tomaszantas (approved):
Looks good to me!
I have added one commit exporting two types - probably issue not related to changes in this PR.

One issue I found is that the user has to hover icon to see tooltip instead of button itself. I'm adding this on Linear: [LAY-1001](https://linear.app/layerfi/issue/LAY-1001/add-support-for-tooltip-on-buttons)

PR #358, opened 2024-10-07
feat: receipt upload ui

tomaszantas:
## Description

Add new UI for receipt upload in Bank Transactions

image

### DIFF
--- a/src/components/BankTransactionsLoader/BankTransactionsLoader.tsx
+++ b/src/components/BankTransactionsLoader/BankTransactionsLoader.tsx
@@ -8,7 +8,7 @@ import { SkeletonTableLoader } from '../SkeletonTableLoader'
 
 export const BankTransactionsLoader = ({
   isLoading = true,
-  showTooltips
+  showTooltips,
 }: {
   isLoading: boolean
   showTooltips: boolean
@@ -47,7 +47,7 @@ export const BankTransactionsLoader = ({
       cols={
         isLoading
           ? [
-              { colSpan: 4, trimLastXRows: 3 },
+              { colSpan: 5, trimLastXRows: 4 },
               { colSpan: 1, parts: 2 },
             ]
           : [
--- a/src/components/ExpandedBankTransactionRow/ExpandedBankTransactionRow.tsx
+++ b/src/components/ExpandedBankTransactionRow/ExpandedBankTransactionRow.tsx
@@ -107,7 +107,6 @@ const validateSplit = (splitData: RowState) => {
   return valid
 }
 
-// @TODO move to global ts
 export interface DocumentWithStatus {
   url?: string
   status: 'pending' | 'uploaded'
@@ -489,8 +488,6 @@ export const ExpandedBankTransactionRow = forwardRef<SaveHandle, Props>(
 
     const className = 'Layer__expanded-bank-transaction-row'
     const shouldHide = !isOpen && isOver
-
-    console.log('receiptUrls', receiptUrls)
 
     return (
       <span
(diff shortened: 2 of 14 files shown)


src/components/ExpandedBankTransactionRow/ExpandedBankTransactionRow.tsx
```diff
@@ -617,38 +674,30 @@ export const ExpandedBankTransactionRow = forwardRef<SaveHandle, Props>(
...
-                          bankTransactionId: bankTransaction.id,
-                          file: file,
-                          documentType: 'RECEIPT',
-                        })
-                        await fetchDocuments()
-                      }}
-                      text='Upload receipt'
+                      secondary
+                      onUpload={onReceiptUpload}
+                      text='Add next receipt'
```
AndrewAcomb, line 698:
```suggestion
                      text='Upload receipt'
```

src/components/ExpandedBankTransactionRow/ExpandedBankTransactionRow.tsx
```diff
@@ -441,14 +477,101 @@ export const ExpandedBankTransactionRow = forwardRef<SaveHandle, Props>(
 
     useEffect(() => {
       const loadDocumentsAndMetadata = async () => {
-        if (showDescriptions) await fetchMetadata()
-        if (showReceiptUploads) await fetchDocuments()
+        if (showDescriptions && bankTransaction?.metadata.memo)
+          setMemoText(bankTransaction?.metadata.memo)
```
AndrewAcomb, line 481: Made this change so it doesn't separately call the metadata endpoint for each transaction on initial load, instead using the metadata field on the bank transaction object.

src/components/BankTransactionsLoader/BankTransactionsLoader.tsx
```diff
@@ -8,7 +8,7 @@ import { SkeletonTableLoader } from '../SkeletonTableLoader'
 
 export const BankTransactionsLoader = ({
   isLoading = true,
-  showTooltips
+  showTooltips,
```
AndrewAcomb, line 11: This would certainly be in a different PR, but I am curious what you think about potentially using [Jotai](https://jotai.org/) at some point, since I did a lot of prop drilling for the tooltips / receipt upload props.

AndrewAcomb:
https://github.com/user-attachments/assets/238f773f-7cc8-44ee-910e-f10c4415942e

When there is an error state, you can't click on any of the actions since that requires you to hover on the receipt (which the actions do not overlap) ^ see video

src/components/BankTransactionsLoader/BankTransactionsLoader.tsx
tomaszantas: Definitely there are good use cases for Jotai. The only two minor concerns I have are:
- NPM package weights a bit, ~427kB... but still worth to use imho
- we will need to be careful with mixing atoms with specific local states and data later sent to API

tomaszantas:
@AndrewAcomb here is a fix for the issue from your video: #362 .
I put this into new PR in case if you have some local changes for this branch.
Feel free to merge if all looks good.

AndrewAcomb (commented):
Approving along with https://github.com/Layer-Fi/layer-react/pull/362 .

src/components/BankTransactionsLoader/BankTransactionsLoader.tsx
AndrewAcomb: We had a brief discussion and agreed we should just move some things into context at this stage instead, but will revisit Jotai if we later have issues with too many re-renders (which we have not thus far).

AndrewAcomb (approved):

PR #359, opened 2024-10-07
open receipt in new tab, add new fields + endpoints

AndrewAcomb:
- Add endpoints for get document and archive document
- Add query param for content-disposition for documents (not actually needed but is an option now)
- Add transaction metadata and list of document IDs to transaction object
- Add function for opening document image in new tab (need to decide how to handle PDFs)

### DIFF
--- a/src/api/layer/bankTransactions.ts
+++ b/src/api/layer/bankTransactions.ts
@@ -7,7 +7,7 @@ import {
 } from '../../types/bank_transactions'
 import { FileMetadata } from '../../types/file_upload'
 import { S3PresignedUrl } from '../../types/general'
-import { get, put, postWithFormData } from './authenticated_http'
+import { get, put, postWithFormData, post } from './authenticated_http'
 
 export type GetBankTransactionsReturn = {
   data?: BankTransaction[]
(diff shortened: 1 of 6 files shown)


src/components/ExpandedBankTransactionRow/ExpandedBankTransactionRow.tsx
```diff
@@ -644,6 +663,7 @@ export const ExpandedBankTransactionRow = forwardRef<SaveHandle, Props>(
                         href={url}
                         target='_blank'
                         rel='noopener noreferrer'
+                        // onClick={openReceiptInNewTab(url, index)}
```
AndrewAcomb, line 666: Kept this commented out since it does not work for PDFs and would technically be a regression but feel free to try locally. The actual solution here will be using the fileType field on the S3 URL ("application/pdf" or "image/png") and varying behavior accordingly

AndrewAcomb: Noting here that I am not actually doing so in this PR because we should get https://github.com/Layer-Fi/layer-react/pull/358 merged first

src/api/layer/bankTransactions.ts
```diff
@@ -118,7 +118,23 @@ export const listBankTransactionDocuments = get<{
...
-    `/v1/businesses/${businessId}/bank-transactions/${bankTransactionId}/documents`,
+    `/v1/businesses/${businessId}/bank-transactions/${bankTransactionId}/documents?content_disposition=INLINE`,
+)
+
+export const getBankTransactionDocument = get<{
+  data: S3PresignedUrl
+  errors: unknown
+}>(
+  ({ businessId, bankTransactionId, documentId }) =>
+    `/v1/businesses/${businessId}/bank-transactions/${bankTransactionId}/documents/${documentId}?content_disposition=INLINE`,
```
AndrewAcomb, line 129: FYI @tomaszantas  , this does not actually do anything since `inline` is the default content disposition for S3. Instead, I have created an example function that opens an image in a new tab.

PR #374, opened 2024-10-17
Project View: Move State in a Context

imkesin:
## Description

This adds `zustand` as a state management solution (we could easily use `jotai` or some other similar solution).


imkesin:
This will be revisited after more work on tags.

PR #375, opened 2024-10-17
Daniel/pnl tag filtering

doneel:


src/views/ProjectProfitability/ProjectProfitability.tsx
```diff
@@ -122,53 +125,14 @@ export const ProjectProfitabilityView = ({
...
-                </Container>
-              </div>
+              <AccountingOverview
+                stringOverrides={{}}
+                tagFilter={tagFilter ? tagFilter : undefined}
+                onTransactionsToReviewClick={() => {
+                  console.log('clicked')
+                }}
+                enableOnboarding={false}
+              />
```
AndrewAcomb, line 135: This works but getting the PNL report to work this way is pretty invasive. In my branch I just use the tagFilter field on the ProfitAndLoss component an dit works for both the ProfitAndLoss.DetailedCharts and ProfitAndLoss.Report

AndrewAcomb: That said I do like the way this looks better for the overview tab, so I suggest that we just go with my approach for the report and yours for the accounting overview

AndrewAcomb (approved):
lgtm, merge when ready

PR #376, opened 2024-10-18
project tweaks

AndrewAcomb:
Remove header for overview, make transactions categorized only, adjust z axis for project dropdown, make download button use tag filters

### DIFF
--- a/src/styles/profit_and_loss_detailed_charts.scss
+++ b/src/styles/profit_and_loss_detailed_charts.scss
@@ -160,6 +160,7 @@ header.Layer__profit-and-loss-detailed-charts__header--tablet {
       white-space: nowrap;
       overflow: hidden;
       text-overflow: ellipsis;
+      z-index: 999;
 
       &.Layer__select__option--is-focused,
       &.Layer__select__option--is-selected,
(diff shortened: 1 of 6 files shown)


src/styles/profit_and_loss_detailed_charts.scss
```diff
@@ -160,6 +160,7 @@ header.Layer__profit-and-loss-detailed-charts__header--tablet {
       white-space: nowrap;
       overflow: hidden;
       text-overflow: ellipsis;
+      z-index: 999;
```
doneel, line 163: Does this cause it to clip the project select dropdown? That was causing problems for me earlier.

doneel: If so, z index 1 might be sufficient

doneel (approved):

src/styles/profit_and_loss_detailed_charts.scss
AndrewAcomb: This was speculative because I can't actually replicate the issue justin is having, so I'll just do 1 for now

PR #381, opened 2024-10-21
feat: receipts forms for mobile

tomaszantas:
## Description

Refactor and better support for receipts + memo notes on mobile views.

## How this has been tested?

image

<br/>

image


src/components/BankTransactionReceipts/BankTransactionReceipts.tsx
```diff
@@ -0,0 +1,130 @@
...
+                : undefined
+            }
+            enableDownload
+            error={url.error}
+            onDelete={() => url.id && archiveDocument(url)}
+          />
+        ))}
+        {!hideUploadButtons &&
+        receiptUrls.length > 0 &&
+        receiptUrls.length < 10 ? (
```
AndrewAcomb, line 120: small nit that we should have a const for MAX_RECEIPT_COUNT instead of directly limiting it to 10 here

AndrewAcomb (approved):

src/components/BankTransactionReceipts/BankTransactionReceipts.tsx
tomaszantas: Good point! thanks for suggestion!

PR #383, opened 2024-10-21
fix css for $

AndrewAcomb:


AndrewAcomb:
Was not actually able to replicate the issue that prompted this

imkesin (approved):
~~It would be nice to have a link to the spot in slack where they communicated this issue.~~

Edit:
It wouldn't make sense to show the message in a public repo.

PR #384, opened 2024-10-21
allow custom date range props from top level

doneel:
Adding in an option (specifiable at any level of component) for custom date ranges.

### DIFF
--- a/src/components/DatePicker/DatePickerOptions.tsx
+++ b/src/components/DatePicker/DatePickerOptions.tsx
@@ -12,15 +12,39 @@ import {
   subYears,
 } from 'date-fns'
 
+export interface CustomDateRange {
+  label: string
+  startDate: Date
+  endDate: Date
+}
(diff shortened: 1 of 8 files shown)


src/components/DatePicker/DatePickerOptions.tsx
```diff
@@ -124,6 +143,7 @@ export const DatePickerOptions = ({
       optionsComponents.push(getOptionComponent(option))
     })
   }
+  optionsComponents.push(...customOptionComponents)
```
AndrewAcomb, line 146: Could we make the custom option components the only ones shown here (remove `this-month`, `last-month`, etc) and just set the default value for custom options to include those such that users can disable `this-month`, or `last-month` or any of the other preset options by excluding them?

AndrewAcomb (approved):

imkesin (approved):

PR #392, opened 2024-10-23
fix panel z idx

AndrewAcomb:
This one was fun to track down. The date picker in the manual journal entry form was being blocked by the sidebar panel.

### DIFF
--- a/src/styles/panel.scss
+++ b/src/styles/panel.scss
@@ -42,7 +42,7 @@
   height: 100%;
   border-left: 1px solid var(--color-base-300);
   border-top-right-radius: var(--border-radius-sm);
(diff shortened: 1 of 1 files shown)


src/styles/panel.scss
```diff
@@ -42,7 +42,7 @@
   height: 100%;
   border-left: 1px solid var(--color-base-300);
   border-top-right-radius: var(--border-radius-sm);
-  z-index: 100;
+  z-index: 11;
```
AndrewAcomb, line 45: I would go lower but the journal entries table header has z-idx = 10. At some point we should reevaluate every z index in our components

imkesin (approved):

PR #393, opened 2024-10-23
fix(build): force target ES2016

imkesin:
## Description

A breaking change was made in [esbuild v0.18.0](https://github.com/evanw/esbuild/blob/main/CHANGELOG-2023.md#0180) that dropped inference of `target` from the supplied tsconfig file.

### DIFF
--- a/bin/build.js
+++ b/bin/build.js
@@ -10,12 +10,17 @@ new Generator({
 }).generate()
 
 const entryPoints = ['src/index.tsx']
+
+/**
+ * @type {import('esbuild').BuildOptions}
+ */
 const sharedConfig = {
   entryPoints,
   bundle: true,
   minify: false,
   sourcemap: true,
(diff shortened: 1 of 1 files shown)


bin/build.js
```diff
@@ -10,12 +10,17 @@ new Generator({
...
+
+/**
+ * @type {import('esbuild').BuildOptions}
+ */
 const sharedConfig = {
   entryPoints,
   bundle: true,
   minify: false,
   sourcemap: true,
   tsconfig: './tsconfig.json',
```
imkesin, line 22: This documentation [here](https://esbuild.github.io/content-types/#tsconfig-json) is a bit outdated. Still, I believe there are other settings in `tsconfig.json` that are relevant to a build.

AndrewAcomb (approved):
lgtm

PR #396, opened 2024-10-28
[Do Not Merge] Issue: Not Using PostCSS

imkesin:
## Description

We are not using PostCSS at all in our component library. There is an illusion of using PostCSS because it is implicitly called through `react-scripts` (which we should also not use) in our demo app.


imkesin:
This is something that I will revisit in the future

PR #400, opened 2024-10-28
fix(pnl-summaries)!: use slot for `TransactionsToReviewCard`

imkesin:
## Description

The `TransactionsToReviewCard` component was not exposed in the public API surface, but was suggested through documentation.

The `ProfitAndLossSummaries` component needs to pivot and scale as unit, so it makes more sense for it to behave as an unordered list of cards.

### Screenshots

Small (Default):
Default_Small_Variant

Large (New):
Large_Variant

### Note

I have not removed the soft border on the `TransactionsToReviewCard` yet, so it looks like a double border. This should've been an internal component, but someone could be using it in the way we've recommended in our docs.


src/components/ProfitAndLossSummaries/ProfitAndLossSummaries.tsx
```diff
@@ -54,149 +56,113 @@ const buildMiniChartData = (scope: Scope, data?: ProfitAndLoss) => {
   return items.slice()
 }
 
+const SECTION_CLASS_NAME = 'Layer__ProfitAndLossSummaries'
+const SECTION_CLASS_NAMES = `${SECTION_CLASS_NAME} Layer__component`
```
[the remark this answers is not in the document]
doneel: We do have existing terms `View` and `Component` - is your concept of `Section` independent from those? I _think_ it would be based on your description that sections can go inside of other sections: sections are just whatever gets publicly exposed, which could be components or views (or something in the future).

src/components/ProfitAndLossSummaries/ProfitAndLossSummaries.tsx
```diff
@@ -1,202 +1,136 @@
...
 type ProfitAndLossSummariesProps = {
-  vertical?: boolean
   actionable?: boolean
+  stringOverrides?: ProfitAndLossSummariesStringOverrides
+  chartColorsList?: string[]
+  slots?: {
+    unstable_AdditionalListItems?: [ReactNode]
+  }
+  variants?: {
+    size: 'sm' | 'lg'
```
doneel, line 30: I would imagine this should be an optional property. In the future I'm sure we'll have use cases for a variant type other than size and we'd want them to be independently specifiable, yes?

src/components/ProfitAndLossSummaries/ProfitAndLossSummaries.tsx
```diff
@@ -54,149 +56,113 @@ const buildMiniChartData = (scope: Scope, data?: ProfitAndLoss) => {
...
+const SECTION_CLASS_NAME = 'Layer__ProfitAndLossSummaries'
+const SECTION_CLASS_NAMES = `${SECTION_CLASS_NAME} Layer__component`
+
 type ProfitAndLossSummariesProps = {
-  vertical?: boolean
   actionable?: boolean
+  stringOverrides?: ProfitAndLossSummariesStringOverrides
+  chartColorsList?: string[]
+  slots?: {
+    unstable_AdditionalListItems?: [ReactNode]
```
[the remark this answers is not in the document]
doneel: The goal here, as I understand it, is to allow us internally to set `additionalListItem` (for use cases like the `<TransactionsToReviewCard />`), but we don't want customers to use this slot. Which makes sense!

I think we could accomplish this by having a very simple wrapper component which _doesn't_ include this optional prop but otherwise delegates entirely to this internal version. Then we only publish/expose the wrapper so customers only see a version where `unstableAdditionalListItems` isn't even a prop.

src/components/ProfitAndLossSummaries/internal/profit_and_loss_summaries_list.scss
```diff
@@ -0,0 +1,42 @@
...
+  box-shadow: 0px 0px 0px 1px var(--color-base-300);
+
+  padding: var(--spacing-4xs);
+
+  &[data-active] {
+    box-shadow: 0px 0px 0px 1px var(--color-base-200);
+    background: var(--color-base-50);
+  }
+
+  &[data-clickable] {
```
doneel, line 16: I appreciate the data attribute pattern here now. Nice improvement!

src/components/ProfitAndLossSummaries/internal/profit_and_loss_summaries_summary.scss
```diff
@@ -0,0 +1,56 @@
...
+  min-block-size: 3rem;
+  grid-template-columns: auto minmax(0, 1fr);
+  column-gap: var(--spacing-xs);
+  row-gap: var(--spacing-3xs);
+  grid-template-areas:
+    'chart heading'
+    'chart amount'
+  ;
+
+  &[data-size="lg"] {
```
doneel, line 12: I like this pattern because I believe to support a new size, we'd just need to add it into the possible prop values and then modify _this one css file_ to support it. Is that right?

src/config/charts.ts
```diff
@@ -19,81 +19,6 @@ export const DEFAULT_CHART_COLOR_TYPE = [
   '#71CC56',
 ]
 
-export const DEFAULT_MINICHART_COLORS = [
```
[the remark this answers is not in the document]
doneel: That makes sense, good drop

src/views/AccountingOverview/AccountingOverview.tsx
```diff
@@ -77,18 +77,22 @@ export const AccountingOverview = ({
...
+              ? [
+                  <TransactionToReviewCard
+                    key='transactions-to-review'
+                    usePnlDateRange={true}
+                    onClick={onTransactionsToReviewClick}
+                  />,
+                ]
+              : undefined,
+          }}
+          variants={{ size: 'lg' }}
```
doneel, line 94: As I understood it, `sm` corresponds to what our components _currently_ default to. Setting this to `lg` would actually make the _default_ much larger, right? 

If I have that right, I'd much rather default to `sm` to preserve the current styling.

doneel (approved):
Looks good - a few items we can discuss async on this pr!

Request: Can you post screenshots of the new styles in this PR so we have those visible for reviewers now and in perpetuity?

src/views/AccountingOverview/AccountingOverview.tsx
doneel: One more important one: unless we expose the `variants` argument on `AccountingOverview`, customers won't actually be able to set this variant as in practice, they're using the `<AccountingOverview />` component to get this whole layout rather than using the `<ProfitAndLoss.Summaries />` where they could actually set the styleing.

src/components/ProfitAndLossSummaries/internal/profit_and_loss_summaries_summary.scss
imkesin: Yes, exactly.

It's non-negotiable that we co-locate our styling. When you change the behavior of one CSS class, it should be extremely obvious that you are changing the appearance of exactly one component.

src/components/ProfitAndLossSummaries/ProfitAndLossSummaries.tsx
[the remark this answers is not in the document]
imkesin: `Component` seems like a wildcard term - it has no indication of it's role in a hierarchy.

Section has an semantic meaning in html (`<section>...</section>`). Sections are supposed to be siblings of each other, and they can be placed within a layout.

It seems like our `View` is a layout that has been populated with several sections. A `View` is the faster way to get something working with our components, but should not - in my opinion - support every opinion of the wrapped sections.

README.md
```diff
@@ -168,26 +168,10 @@ import { ProfitAndLoss } from "@layerfi/components";
...
 </ProfitAndLoss>
 ```
 
-Props:
-
-- `actionable`: enables or disables whether clicking the revenue & expense charts open the P&L sidebar view.
-- `vertical`: changes the card layout to be vertically stacked instead of horizontal
-- `revenueLabel`: specifiable label for revenue for uses where you prefer 'income' or other another term.
-
-Note that the `<TransactionToReviewCard>` is a separate component, but is meant to be optionally bundled with the summary cards. As with the onboarding component, this component has one primary prop: `onTransactionsToReviewClick` should be a function which navigates to the bank transactions to review page. For example, if the bank transaction categorization page lives on `/account/bank-transactions` within your app:
```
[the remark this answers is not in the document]
imkesin: I'll add back `actionable` because it isn't deprecated.

src/components/ProfitAndLossSummaries/ProfitAndLossSummaries.tsx
```diff
@@ -54,149 +56,121 @@ const buildMiniChartData = (scope: Scope, data?: ProfitAndLoss) => {
...
     isLoading,
     setSidebarScope,
     sidebarScope,
   } = useContext(PNL.Context)
 
+  /*
+   * What is going on here? Why is this not explicitly an array type?
+   *
+   * This introduces unnecessary `any` types
+   */
```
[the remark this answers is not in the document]
imkesin: @ArkuVonSymfon - any context on this?

src/components/ProfitAndLossSummaries/ProfitAndLossSummaries.tsx
imkesin: Converted to an optional property.

src/views/AccountingOverview/internal/TransactionsToReview.tsx
```diff
@@ -0,0 +1,110 @@
...
+  onClick?: () => void
+  usePnlDateRange?: boolean
+  tagFilter?: {
+    key: string
+    values: string[]
+  }
+  variants?: Variants
+}
+
+export function TransactionsToReview({
```
imkesin, line 29: This file was renamed, so it appears as completely new.

The goal here was to change as little as possible.

doneel (approved):

PR #402, opened 2024-10-30
Daniel/lg summary style suggestions

doneel:
# Transactions to Review Title
1. I've made it share the styling as the headers of the other cards using the same CSS class. 
2. I've made the vertical spacing depend on the `variant.size` so it looks more like the spacing on the other cards.

#### Before
image

#### After
image

# MiniCharts
I've made the minicharts scale up in size as well. 
#### Side by side comparison
image

### DIFF
--- a/src/views/AccountingOverview/internal/TransactionsToReview.tsx
+++ b/src/views/AccountingOverview/internal/TransactionsToReview.tsx
@@ -1,20 +1,22 @@
-import React, { useContext, useEffect, useState } from 'react'
+import React, { useContext, useEffect, useMemo, useState } from 'react'
 import { Badge } from '../../../components/Badge'
 import { BadgeSize, BadgeVariant } from '../../../components/Badge/Badge'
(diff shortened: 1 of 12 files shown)


src/views/AccountingOverview/internal/TransactionsToReview.tsx
```diff
@@ -65,12 +67,22 @@ export function TransactionsToReview({
...
+  }
   return (
     <div onClick={onClick} className={CLASS_NAME}>
-      <VStack gap='3xs' align='start'>
-        <Text size={size === 'lg' ? TextSize.lg : TextSize.sm}>
+      <VStack gap={verticalGap} align='start'>
+        <h3 className={HEADING_CLASS_NAME} {...labelDataProperties}>
           Transactions to review
-        </Text>
+        </h3>
```
imkesin, line 85: Suggestion - make this `H3` its own component.

doneel: done

PR #404, opened 2024-10-31
chore: react hooks eslint

imkesin:
## Description

The [react hooks plugin](https://www.npmjs.com/package/eslint-plugin-react-hooks) is more valuable and should be added at some point.

This PR gets to the point where all remaining issues are related to usage of `any`. These are much more difficult to resolve.


src/components/Button/CloseButton.tsx
```diff
@@ -2,14 +2,12 @@ import React, { ButtonHTMLAttributes } from 'react'
 import CloseIcon from '../../icons/CloseIcon'
 import classNames from 'classnames'
 
-export interface CloseButtonProps
-  extends ButtonHTMLAttributes<HTMLButtonElement> {
+type CloseButtonProps = {
   textOnly?: boolean
-}
+} & Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'children'>
```
tomaszantas, line 7: FYI: I had strange behavior when Prettier was enabled in the VSCode. After reloading Prettier, it gave other strange behavior. But I believe that we no longer use Prettier

tomaszantas: I believe same change could be applied to `BackButton`

tomaszantas (approved):
LGTM!

src/components/Button/CloseButton.tsx
imkesin: Thank you for spotting that. I'll go change that component.

PR #405, opened 2024-10-31
fix: bank transactions - monthly view - use date range in api queries

tomaszantas:
## Description

Use filter's `dateRange` when bank transactions is shown in the monthly view.

## How this has been tested?

The history could be cut off after few months because only first 200 records were fetched.
For the `8a...4ace` business, the history of categorized cut off on March, so February was empty:

image

After fix:

image


src/hooks/useIsVisible/index.ts
```diff
@@ -0,0 +1 @@
+export { useIsVisible } from './useIsVisible'
```
imkesin, line 1: Nit: I'd like us to [stop using barrel files](https://tkdodo.eu/blog/please-stop-using-barrel-files).

src/hooks/useBankTransactions/useBankTransactions.tsx
```diff
@@ -377,6 +394,12 @@ export const useBankTransactions: UseBankTransactions = params => {
       })
   }
 
+  useEffect(() => {
+    if (monthlyView && lastMetadata?.pagination?.has_more) {
+      fetchMore()
+    }
+  }, [lastMetadata])
+
```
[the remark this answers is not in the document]
tomaszantas: Yeah, I agree and I already changed it. In the new approach, I removed this `useEffect` and new pages are being fetched when user scrolls to the end of the list.

src/hooks/useIsVisible/index.ts
tomaszantas: Make sense to me. Which approach do you prefer:
1. `utils/useIsVisible/useIsVisible.ts`,
2. `utils/useIsVisible/index.ts`,
3. `utils/useIsVisible.ts`?

I would go with the last one, and when we need to split logic into more files, then [2]. What do you think?
I would make such changes in batch after merging this and eslint related pull requests.

imkesin: I like the third approach. 

As soon as we get an unreasonable number of files at the same level, then we can group a few into a folder.

imkesin (approved):

PR #408, opened 2024-11-06
hide receipt column if not in active use

doneel:
We take up a non-trivial amount of horizontal table space for the receipt icon column. On tighter screens or with a constrained width like some customers enforce, it's noticeable.

This PR adds CSS states to dynamically hide or show this column. It's always rendered, but with zero width _unless both_
1. The `showReceiptUploads` feature flag is on.
2. One row in the current view actually _has_ a receipt upload.

### Before
image

### After
image

You'll notice I also just reduced the space to the left of the `amount` column (unrelated to receipts) because it looks pretty excessive.

**Request for input**: I'd be open to _just_ using the feature flag to determine if we show the column, i.e. if receipts are enabled but you haven't uploaded any, we would _still_ show the column. It means less space in this case, but it leads to a more consistent experience where you don't have an extra column appear.


imkesin (approved):
These types of customizations are going to easily get out of control.

When we next want to offer more power, we should migrate to [TanStack table](https://tanstack.com/table/latest).

tomaszantas (approved):

PR #410, opened 2024-11-07
refactor!: add a dev-mode into the build flow

imkesin:
## Description

I have been extensively testing this locally... but I would appreciate at least one other person trying it on their machine. I would advise deleting the `/dist` directory so that there is an obvious clean slate.

### DIFF
--- a/bin/build.js
+++ b/bin/build.js
@@ -1,43 +1,108 @@
-const { build } = require('esbuild')
+const { build, context } = require('esbuild')
 const { dependencies, peerDependencies } = require('../package.json')
 const { Generator } = require('npm-dts')
 const { sassPlugin } = require('esbuild-sass-plugin')
+const { rename, rm } = require('node:fs/promises')
+
+const OUT_DIR = 'dist'
+
+const args = process.argv.slice(2)
+const watchModeEnabled = args.includes('--watch')
 
 new Generator({
   entry: 'index.tsx',
-  output: 'dist/index.d.ts',
+  output: `${OUT_DIR}/index.d.ts`,
   tsc: '-p ./tsconfig.json',
-}).generate()
+},
+  null,
+  !watchModeEnabled
+)
+.generate()
+.catch(() => {
+  throw new Error(
+    [
+      "Failed to generate types:",
+      "  - Try running `npx npm-dts generate -L debug` for more information",
+    ].join('\n'),
+  )
+})
 
-const entryPoints = ['src/index.tsx']
-
-/**
- * @type {import('esbuild').BuildOptions}
- */
+/** @type {import('esbuild').BuildOptions} */
 const sharedConfig = {
-  entryPoints,
+  entryPoints: [ 'src/index.tsx' ],
   bundle: true,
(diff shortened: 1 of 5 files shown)


bin/build.js
```diff
@@ -1,43 +1,108 @@
...
+  return Promise.all(
+    [
+      rename(`${baseDir}/styles/index.css`, `${OUT_DIR}/index.css`),
+      withMap
+        ? rename(`${baseDir}/styles/index.css.map`, `${OUT_DIR}/index.css.map`)
+        : null
+    ].filter(Boolean)
+  )
+  .then(() => rm(`${baseDir}/styles`, { recursive: true }))
+}
```
doneel, line 64: Am I understanding correctly that the way we can make it easier to import the CSS is to add to build step to literally just move the file?

bin/build.js
```diff
@@ -1,43 +1,108 @@
...
+    ...esmConfig,
+    target: 'esnext',
+    minify: false,
+    sourcemap: true,
+    plugins: [
+      ...esmConfig.plugins,
+      {
+        name: 'rebuild-notify',
+        setup(build) {
+          build.onEnd(() => {
```
doneel, line 91: Are the build settings/context different in this hot reload build versus the normal full build? I'm not sure I'm reading it correctly, I'm just looking for a way to reduce the differences between the two environments

src/components/DatePicker/DatePicker.tsx
```diff
@@ -310,7 +322,7 @@ export const DatePicker = ({
         onFocus={e => (e.target.readOnly = true)}
         onInputClick={() => {
           if (pickerRef.current && !isDesktop) {
-            pickerRef.current.setOpen(!pickerRef.current.isCalendarOpen)
+            pickerRef.current.setOpen(!pickerRef.current.isCalendarOpen())
```
doneel, line 325: How was this working before? Either it was broken or I'm missing some knowledge.

doneel (approved):
Looks exciting a couple clarification questions, but merge when you're ready.

bin/build.js
imkesin: Similar to our exported types, our CSS is not specific to ESM or CJS.  It happens to get initially nested inside a certain sub-folder, so I explicitly move it out of there.

The change in `package.json` (see `"exports" and "style"`) should make it so our css files end-up in a standard location.

src/components/DatePicker/DatePicker.tsx
imkesin: The types for `react-datepicker` are awful. They only ported to a typescript approach in the latest version 7.X (and we're on 6.9 with an external types library).

It is actually a method, not a property, so it should be called.

This case can't even be reached on a mobile view due to some dark magic with the `onFocus` handler above ([reference to source code](https://github.com/Hacker0x01/react-datepicker/blob/02ce53f2c50ae77b177da1021b6e7f2e6faaaf0f/src/index.jsx#L484)). Still, the end-result is absolutely fine.

PR #414, opened 2024-11-13
fix: auth hook should poll

imkesin:
## Description

While developing, I ran into a bug where my authentication token has gone stale. I have created a new hook to request a token periodically.

This hook is also one of the key steps on the path to refactoring our data-fetching logic. Most of the state should not be copied over into contexts... though I have left this behavior in-place for now.


src/api/layer/bankTransactions.ts
```diff
@@ -132,15 +132,15 @@ export const getBankTransactionDocument = get<{
...
+  data: Record<never, never>
   errors: unknown
 }>(
   ({ businessId, bankTransactionId, documentId }) =>
     `/v1/businesses/${businessId}/bank-transactions/${bankTransactionId}/documents/${documentId}/archive`,
 )
 
 export const uploadBankTransactionDocument =
-  (baseUrl: string, accessToken: string) =>
+  (baseUrl: string, accessToken?: string) =>
```
bryceAebi, line 143: should this function not be called if accessToken is undefined? same questien in tasks.ts

bryceAebi (approved):
Looks much cleaner!

tomaszantas (approved):

PR #416, opened 2024-11-14
refactor: move environment-specific input into own provider

imkesin:
## Description

This look worse in the short-term, but is part of a larger effort.


tomaszantas (approved):
Looks good to me and perhaps would make also sens to move `theme` into another context.

PR #417, opened 2024-11-14
Fix QB OAuth bug

bryceAebi:
When QB redirects back to the Layer app, there was a race condition between fetching the auth token and fetching some quickbooks data. This ensures we have the auth token before querying the quickbooks data

### DIFF
--- a/src/hooks/useQuickbooks/useQuickbooks.ts
+++ b/src/hooks/useQuickbooks/useQuickbooks.ts
@@ -1,88 +1,90 @@
-import { useEffect, useRef, useState } from 'react'
-import { Layer } from '../../api/layer'
-import { useLayerContext } from '../../contexts/LayerContext'
+import { useEffect, useRef, useState } from "react";
+import { Layer } from "../../api/layer";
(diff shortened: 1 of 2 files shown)


src/hooks/useQuickbooks/useQuickbooks.ts
```diff
@@ -1,94 +1,96 @@
...
+  }, [isSyncingFromQuickbooks]);
 
   // Determine whether there exists an active Quickbooks connection or not
   useEffect(() => {
-    fetchQuickbooksConnectionStatus()
-  }, [])
+    if (auth?.access_token) {
+      fetchQuickbooksConnectionStatus();
+    }
+  }, [auth?.access_token]);
```
bryceAebi, line 41: this is the change

imkesin (approved):
Do I have permission to refactor this into a SWR-based flow at some point in the future?

bryceAebi:
> Do I have permission to refactor this into a SWR-based flow at some point in the future?

@imkesin feel free to change/fix anything. I just took the shortest path to removing the error while I'm testing QB things

PR #420, opened 2024-11-15
chore: remove many uses of explicit `any`

imkesin:
## Description

This is part of an effort to get us down to a few linter errors as possible. Once we can get it to zero, then we can fail the production build

### DIFF
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -14,9 +14,11 @@ import classNames from 'classnames'
 /**
  * @see https://github.com/Hacker0x01/react-datepicker/issues/1333#issuecomment-2363284612
  */
+/* eslint-disable @typescript-eslint/no-explicit-any */
 const ReactDatePicker = (((RDP.default as any).default as any)
   || (RDP.default as any)
   || (RDP as any)) as typeof RDP.default
+/* eslint-enable @typescript-eslint/no-explicit-any */
 
 interface DatePickerProps {
   mode: DatePickerMode
--- a/src/components/ProfitAndLossChart/Indicator.tsx
+++ b/src/components/ProfitAndLossChart/Indicator.tsx
@@ -25,10 +25,6 @@ export const Indicator = ({
   setCustomCursorSize,
   viewBox = {},
 }: Props) => {
-  if (!className?.match(/selected/)) {
-    return null
-  }
-
   const [opacityIndicator, setOpacityIndicator] = useState(0)
 
   const { x: animateTo = 0, width = 0 } =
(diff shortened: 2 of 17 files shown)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -14,9 +14,11 @@ import classNames from 'classnames'
 /**
  * @see https://github.com/Hacker0x01/react-datepicker/issues/1333#issuecomment-2363284612
  */
+/* eslint-disable @typescript-eslint/no-explicit-any */
 const ReactDatePicker = (((RDP.default as any).default as any) ||
   (RDP.default as any) ||
   (RDP as any)) as typeof RDP.default
+/* eslint-enable @typescript-eslint/no-explicit-any */
```
imkesin, line 21: This is one of the few places where it is not obvious how to remove `any`.

I think we need to eventually move away from / upgrade `ReactDatePicker`. The version of the package we're on has strange types because the whole thing was written in JS and augmented by an @types package.

src/components/ProfitAndLossChart/Indicator.tsx
```diff
@@ -52,6 +48,10 @@ export const Indicator = ({
     }, 200)
   }, [animateTo])
 
+  if (!className?.match(/selected/)) {
+    return null
+  }
```
imkesin, line 53: This is one of the last `rules-of-hooks` changes. Next step is to audit dependency arrays for hook.

src/components/ProfitAndLossChart/ProfitAndLossChart.tsx
```diff
@@ -508,16 +508,27 @@ export const ProfitAndLossChart = ({
     )
   }
 
-  const CustomizedYTick = (props: any) => {
+  const CustomizedYTick = (
+    {
+      verticalAnchor: _verticalAnchor,
+      visibleTicksCount: _visibleTicksCount,
+      tickFormatter: _tickFormatter,
```
imkesin, line 515: This removes a bunch of console warnings for spreading unexpected props.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -36,6 +36,8 @@ export const TooltipTrigger = forwardRef<
   HTMLProps<HTMLElement> & { asChild?: boolean }
 >(function TooltipTrigger({ children, asChild = false, ...props }, propRef) {
   const context = useTooltipContext()
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
imkesin, line 40: This is the second instance where I don't have a good solution to factor-out `any`.

tomaszantas (approved):

src/components/Tooltip/Tooltip.tsx
imkesin: Super clever ❤️

PR #422, opened 2024-11-19
fix: reset modal when polling completes

imkesin:
## Description

This is intended to avoid infinite loading in the (rare) case without any results.

### Eng Note

The class-based structure is a bit rushed - this is a design I'm familiar with... but am sure we can tweak when we see another use-case.


imkesin:
This isn't the right structure either - it seems like I shouldn't need to poll at all

PR #428, opened 2024-11-20
chore: next version of ESLint

imkesin:
## Description

The next version of ESLint requires moving to the "flat config." This is part of the larger effort to get our repo to a state where we can run assertions in CI.

Almost all of the ~1.4K of the found errors are stylistic and auto-fixable. Those are going to require a massive diff, so I have not included them here (to manage bloat).


tomaszantas (approved):
LGTM!

PR #436, opened 2024-11-26
feat: download button for casflow statement and balance sheet

imkesin:
## Description

This change adds a button at the top of the balance sheet and cashflow reports to download them in CSV form. The goal is to reach parity with the Income Statement.

### Screenshots

Balance Sheet w/ Download Button
Cashflow Statement w/ Download Button

### DIFF
--- a/src/components/StatementOfCashFlow/StatementOfCashFlow.tsx
+++ b/src/components/StatementOfCashFlow/StatementOfCashFlow.tsx
@@ -16,6 +16,7 @@ import { StatementOfCashFlowTableStringOverrides } from '../StatementOfCashFlowT
 import { View } from '../View'
 import { STATEMENT_OF_CASH_FLOW_ROWS } from './constants'
 import { endOfMonth, startOfDay, startOfMonth, subWeeks } from 'date-fns'
+import { CashflowStatementDownloadButton } from './download/CashflowStatementDownloadButton'
 
 const COMPONENT_NAME = 'statement-of-cash-flow'
 
@@ -116,6 +117,12 @@ const StatementOfCashFlowView = ({
           <Header>
             <HeaderRow>
               <HeaderCol>{datePicker}</HeaderCol>
+              <HeaderCol>
+                <CashflowStatementDownloadButton
+                  startDate={startDate}
+                  endDate={endDate}
+                />
+              </HeaderCol>
             </HeaderRow>
           </Header>
         }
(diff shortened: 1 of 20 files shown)


src/components/utility/InvisibleDownload.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const triggerInvisibleDownload = useCallback((options: { url: string }) => {
+    invisibleDownloadRef.current?.trigger(options)
+  }, [])
+
+  return { invisibleDownloadRef, triggerInvisibleDownload }
+}
+
+const CLASS_NAME = 'Layer__InvisibleDownload'
+
+const InvisibleDownload = forwardRef<InvisibleDownloadHandle>((_props, ref) => {
```
imkesin, line 20: This is an intentional departure from `location.href =` 

I believe that this approach is (marginally) more defensible, since we're using the intended pathway for browser downloads. The former is a redirect that happens to produce a download.

src/components/StatementOfCashFlow/StatementOfCashFlow.tsx
```diff
@@ -116,6 +117,12 @@ const StatementOfCashFlowView = ({
           <Header>
             <HeaderRow>
               <HeaderCol>{datePicker}</HeaderCol>
+              <HeaderCol>
+                <CashflowStatementDownloadButton
+                  startDate={startDate}
+                  endDate={endDate}
+                />
```
tomaszantas, line 124: NIT: you can add the following to have icon-only download button (like on BalanceSheet)

```suggestion
import { useElementViewSize } from '../../hooks/useElementViewSize'
import { View as ViewType } from '../../types/general'

  const [view, setView] = useState<ViewType>('desktop')
  const containerRef = useElementViewSize<HTMLDivElement>(newView =>
    setView(newView),
  )

<View
  (...)
  ref={containerRef}
  >

                <CashflowStatementDownloadButton
                  startDate={startDate}
                  endDate={endDate}
                  iconOnly={view === 'mobile'}
                />
```

image

tomaszantas (approved):
Looks good! I left one minor suggestion to show download button as simple icon button in StatementOfCashflow

src/components/StatementOfCashFlow/StatementOfCashFlow.tsx
imkesin: I appreciate the catch.

I think we should be able to do even-better in the long-term with CSS-only approaches (as we still have a bit of flicker / layout shift).

PR #440, opened 2024-12-04
Add PNL pdf generator component

bryceAebi:
Screenshot 2024-12-06 at 11 03 44 AM
Screenshot 2024-12-06 at 11 03 54 AM

Screenshot 2024-12-06 at 11 04 08 AM

Screenshot 2024-12-06 at 11 04 32 AM
Screenshot 2024-12-06 at 11 04 40 AM


src/styles/fonts.scss
```diff
@@ -0,0 +1,13 @@
+@font-face {
+  font-family: 'Inter';
+  src: url('../assets/fonts/Inter/Inter-Italic-VariableFont_opsz,wght.ttf') format('truetype');
+  font-weight: 100 900;
+  font-style: italic;
+}
```
bryceAebi, line 6: PDF generation library would not allow using a remote css file. We should have these downloaded anyway so we don't depend on a font server

imkesin:
This work does not have a direct path to be merged / integrated into the components.

We will revisit the PDF generation effort in the future.

PR #451, opened 2024-12-12
fix: use discretion to resolve eslint errors

imkesin:
## Description

I've attempted to manually resolve the last of the `eslint` errors.

### DIFF
--- a/test/mocks/styleMock.js
+++ b/test/mocks/styleMock.js
@@ -1,1 +1,0 @@
-module.exports = {}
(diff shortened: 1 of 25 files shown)


test/mocks/styleMock.js
imkesin, line 1: This has intentionally been deleted.

We eventually need to revisit our testing strategy.

imkesin:
This has drifted and is no longer relevant.

PR #463, opened 2025-01-08
Revisions: Account Balance Input

imkesin:
## Description

I want to use a separate PR to highlight a few of the things I want to change with the base branch.

### 1. Added type-checked eslint to specifically highlight areas where we are not handling promises correctly

When `onSuccess` is marked as `Awaitable<void>`, I want to have it awaited so that modals do not close until all of the data is properly refreshed.

This change necessitated a few tweaks to `tsconfig.json`, prompted removing all of our unused test files, etc.

### 2. Removed both the success banner and the separate component that also includes confirmation

These are not necessary today, so I want to cut down on the surface area of extra props.

(I think the success banner looks cool, but would strongly prefer to close the modal immediately on success)

### 3. Pushed the mutations intentionally into `useSWRMutation`

SWR will give us loading states and error values (if we request it).

### 4. Tweaked the form default state

I believe it is possible to force default dates and a default amount of $0.00; any changes will still result in selecting another value. I don't think we need error states in this case (though this is good thinking that should be deployed in other cases).

### 5. Adjusted API types

The base PR isn't anticipating a field in the `notifications` array; it was also trying to send a numeric (cents) value as a string.

### 6. Made a few minor changes that weren't resolved after the first CR

The most notable one is the CSS sizing applied to the `Modal` low-level component.


tsconfig.json
imkesin, line 1: I am going to re-work our `tsconfig.json` file at some point.

tsconfig.json
```diff
@@ -11,9 +11,9 @@
     // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
 
     /* Language and Environment */
-    "target": "es2016" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
+    "target": "es2022" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
```
tomaszantas, line 14: IIRC, some client was using older version in their application, and hit some issues when newer version was here. I agree on using something more recent, but we may need to double-check if that won't break anything on the client side.

src/hooks/useQuickbooks/useQuickbooks.ts
```diff
@@ -22,17 +22,18 @@ export const useQuickbooks: UseQuickbooks = () => {
   const [isSyncingFromQuickbooks, setIsSyncingFromQuickbooks] =
     useState<boolean>(false)
   const [quickbooksIsLinked, setQuickbooksIsLinked] = useState<boolean | null>(
-    null
+    null,
   )
-  const syncStatusIntervalRef = useRef<NodeJS.Timeout | null>(null)
+  const syncStatusIntervalRef = useRef<number | null>(null)
```
imkesin, line 27: This is the most notable change in the file - we can't be using Node types when writing browser code.

src/components/StatementOfCashFlow/download/useCashflowStatementDownload.ts
imkesin, line 1: I accidentally pressed save here - no functional changes

src/components/LinkedAccounts/OpeningBalanceModal/OpeningBalanceModal.tsx
```diff
@@ -30,82 +30,61 @@ function LinkedAccountsOpeningBalanceModalContent({
...
+        },
+      ]),
+    ),
+  )
 
-  const [formsData, setFormsData] = useState<AccountFormBoxData[]>(accounts.map(item => (
+  const {
+    trigger,
+    isMutating,
+  } = useBulkSetOpeningBalance(
```
imkesin, line 52: @tomaszantas - If you wanted to handle an error related to the request, then we would unbox it here.

src/components/HoverMenu/HoverMenu.tsx
```diff
@@ -1,12 +1,13 @@
...
 
-export interface HoverMenuProps {
+export type HoverMenuProps = {
   children: ReactNode
-  config: {
+  config: Array<{
     name: string
-    action: () => void
-  }[]
+    action: () => Awaitable<void>
```
tomaszantas, line 9: This makes sense to me, but it breaks in few places, for instance: `LinkedAccountsInactive.tsx`:

```bash
$ npm run typecheck

src/components/LinkedAccountsInactive/LinkedAccountsInactive.tsx:17:16 - error TS2322: Type '{ name: string; action: () => (source: AccountSource, sourceId: string) => void; }[]' is not assignable to type '{ name: string; action: () => Awaitable<void>; }[]'.
  Type '{ name: string; action: () => (source: AccountSource, sourceId: string) => void; }' is not assignable to type '{ name: string; action: () => Awaitable<void>; }'.
    The types returned by 'action()' are incompatible between these types.
      Type '(source: AccountSource, sourceId: string) => void' is not assignable to type 'Awaitable<void>'.

17     <HoverMenu config={hoverActions}>
                  ~~~~~~

  src/components/HoverMenu/HoverMenu.tsx:7:3
    7   config: Array<{
        ~~~~~~
    The expected type comes from property 'config' which is declared here on type 'IntrinsicAttributes & HoverMenuProps'


Found 1 error in src/components/LinkedAccountsInactive/LinkedAccountsInactive.tsx:17
```

src/components/LinkedAccounts/AccountFormBox/AccountFormBox.tsx
```diff
@@ -1,55 +1,38 @@
...
-import { isEqual, startOfDay } from 'date-fns'
-import classNames from 'classnames'
-import { ErrorType } from '../OpeningBalanceModal/useUpdateOpeningBalanceAndDate'
-import CheckCircle from '../../../icons/CheckCircle'
+import { isEqual } from 'date-fns'
+import { dollarsToCents, centsToDollars } from '../../../models/Money'
+
+function centsToDollarsWithoutCommas(cents: number = NaN): string {
+  return centsToDollars(cents).replaceAll(',', '')
+}
```
tomaszantas, line 12: I would move it to the `../../../models/Money` and called ie. `centsToDollarsNumber`. Then the function could be simply:

```
isNaN(cents) ? NaN : Math.round(cents * 100)
```

tomaszantas (commented):
Due to TS issue, couldn't run it locally, and I have hesitation on removing validation, API error handling and success state. Let's discuss on today's call

src/components/LinkedAccounts/AccountFormBox/AccountFormBox.tsx
imkesin: The function you're suggesting here has a slightly different meaning. I'll move it out utility file.

src/components/HoverMenu/HoverMenu.tsx
imkesin: This entire file is unnecessary, so I nuked it.

src/components/LinkedAccounts/OpeningBalanceModal/OpeningBalanceModal.tsx
```diff
@@ -30,82 +30,61 @@ function LinkedAccountsOpeningBalanceModalContent({
...
   const { refetchAccounts } = useLinkedAccounts()
 
-  // Mark if any data has been successfully saved with API
-  // so the refetchAccounts should be called on onClose
-  const [touched, setTouched] = useState(false)
+  const [formsData, setFormsData] = useState<Record<
+    string,
+    AccountFormBoxData & { account: LinkedAccount }>
+  >(
+    () => Object.fromEntries(
```
tomaszantas, line 37: NIT: I have some hesitation on using maps in form states instead of lists in overall. Maps make it easier to access specific record through the key. But in some case, map can be sorted alphabetically by the key during some transforms. Also adding new record in forms can be troublesome since we have to provide `id` that will be generate on the API. However, I don't think that any of this is a case here, so we can keep this that way if it aligns better with Confirm/Exclude.

src/components/ui/Checkbox/checkbox.scss
```diff
@@ -2,6 +2,7 @@
   display: grid;
   grid-template-columns: auto minmax(0, 1fr);
   align-items: center;
+  gap: var(--spacing-xs);
```
tomaszantas, line 5: Checkbox produces unnecessary spacing if children is not provided

image

src/components/LinkedAccounts/AccountFormBox/AccountFormBox.tsx
```diff
@@ -1,55 +1,35 @@
...
+import { dollarsToCents, centsToDollarsWithoutCommas } from '../../../models/Money'
 
 export type AccountFormBoxData = {
-  account: LinkedAccount
-  isConfirmed: boolean
-  openingDate?: Date
-  openingBalance?: string
-  saved?: boolean
+  openingDate: Date
+  openingBalance: number
```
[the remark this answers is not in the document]
imkesin: This is a good point.

I've put up another commit that pushes this to start with a null value.

src/components/LinkedAccounts/OpeningBalanceModal/OpeningBalanceModal.tsx
imkesin: I would like us to revisit tracking separate errors (e.g. one out of 3 failed) when we come back to onboarding.

src/components/ui/Modal/modal.scss
```diff
@@ -22,6 +22,14 @@
 }
 
 .Layer__Modal {
+  inline-size: min(24rem, 90dvi);
+  block-size: min(24rem, 90dvb);
+
+  &[data-size=lg] {
+    inline-size: min(36rem, 90dvi);
+    block-size: min(36rem, 90dvb);
+  }
```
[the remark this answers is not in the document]
imkesin: Can we do `block-size: auto` (and  `max-block-size: 90dvb`) applied to `Layer__Modal`?

src/components/LinkedAccounts/OpeningBalanceModal/OpeningBalanceModal.tsx
imkesin: I agree and will fix this in the future, especially for cases when we need to re-order elements.

src/components/LinkedAccounts/AccountFormBox/accountFormBox.scss
```diff
@@ -10,10 +10,6 @@
     transition: all 0.25s ease-in-out;
     transition-delay: 0.8s;
     max-height: 280px;
-
-    &.Layer__caobfb--confirmed {
-        background-color: var(--color-base-50);
-    }
```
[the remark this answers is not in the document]
imkesin: (Note to self - only applies if we want the Checkbox in this PR)

imkesin:
Closed in favor of [this revision PR from Tom](https://github.com/Layer-Fi/layer-react/pull/473).

imkesin:
Closed in favor of [this revision PR from Tom](https://github.com/Layer-Fi/layer-react/pull/473).

PR #465, opened 2025-01-10
[LAY-1128] feat: unify all major date pickers through global state

imkesin:
## Description

**Linear**: LAY-1128

The three major reports - P&L, Balance Sheet, and Cash Flow - now share common date state.

After a user adjusts the date range for viewing one report, they will preserve this same date range for another report. Since the balance sheet is a point in time, it is simply the end of the defined range.

## Limitations + Known Issues

- The Profit and Loss Comparison does not correctly handle non-month time periods. This is already an issue and will be attacked in follow-on work.

- The Profit and Loss Chart still displays reasonably, but the selected range outline _only_ highlights the first month in the range. I don't think it's reasonable to fold-in even more changes to manage that chart in special situations.

### DIFF
--- a/src/components/BalanceSheet/BalanceSheet.tsx
+++ b/src/components/BalanceSheet/BalanceSheet.tsx
@@ -3,7 +3,7 @@ import { BalanceSheetContext } from '../../contexts/BalanceSheetContext'
 import { TableProvider } from '../../contexts/TableContext'
 import { useBalanceSheet } from '../../hooks/useBalanceSheet'
 import { useElementViewSize } from '../../hooks/useElementViewSize/useElementViewSize'
-import { BalanceSheetDatePicker } from '../BalanceSheetDatePicker'
+import { BalanceSheetDatePicker } from '../BalanceSheetDatePicker/BalanceSheetDatePicker'
 import { BalanceSheetExpandAllButton } from '../BalanceSheetExpandAllButton'
 import { BalanceSheetTable } from '../BalanceSheetTable'
 import { BalanceSheetTableStringOverrides } from '../BalanceSheetTable/BalanceSheetTable'
@@ -12,8 +12,8 @@ import { Header, HeaderCol, HeaderRow } from '../Header'
 import { Loader } from '../Loader'
 import { View } from '../View'
 import { BALANCE_SHEET_ROWS } from './constants'
-import { format, parse, startOfDay } from 'date-fns'
 import { BalanceSheetDownloadButton } from './download/BalanceSheetDownloadButton'
+import { useGlobalDate } from '../../providers/GlobalDateStore/GlobalDateStoreProvider'
 
 export interface BalanceSheetStringOverrides {
   balanceSheetTable?: BalanceSheetTableStringOverrides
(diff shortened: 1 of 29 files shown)


src/components/BalanceSheet/BalanceSheet.tsx
```diff
@@ -51,43 +51,22 @@ const BalanceSheetView = ({
   asWidget = false,
   stringOverrides,
 }: BalanceSheetViewProps) => {
-  const [effectiveDate, setEffectiveDate] = useState(startOfDay(new Date()))
-  const { data, isLoading, refetch } = useBalanceSheet(effectiveDate)
+  const { date: effectiveDate } = useGlobalDate()
```
tomaszantas, line 54: Sanity-check: Is it obligatory that all components will always use/sync with the global date? It definitely simplifies implementation and maintenance.

imkesin: For now, the three main reports (and visualizations) will all share the global state.

I didn't drill into the `DatePicker` itself, so it is still possible to "break free."

src/components/DatePicker/DatePicker.tsx
```diff
@@ -384,17 +329,19 @@ export const DatePicker = ({
...
-            allowedModes={allowedModes ?? [mode]}
-            onChangeMode={onChangeModeInternal}
-          />
-        )}
-        {mode === 'dayRangePicker' && (
+        {(ModeSelector && pickerMode !== 'timePicker')
+          ? (
+            <ModeSelector
+              mode={pickerMode}
+              allowedModes={allowedModes ?? [pickerMode]}
```
doneel, line 336: Does it make sense to have a mode selector with a single allowed mode? Not obvious to me why we'd render the mode selector if there are less than 2 allowed modes.

I see this pre-dated this PR, so ignore if it feels like a side convo

src/components/DatePicker/DatePicker.tsx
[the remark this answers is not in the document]
doneel: I read this comment and still got confused in a few spots - can we rename `mode` to be `displayMode`? I found your description much easier to understand. I don't think this is an externalized prop name we need to be backwards compatible but if you have reason to believe so, even renaming it internally within this component would be helpful IMO.

src/hooks/useProfitAndLoss/useProfitAndLoss.tsx
```diff
@@ -266,9 +223,8 @@ export const useProfitAndLoss: UseProfitAndLoss = (
     isLoading,
     isValidating,
     error: error,
-    dateRange: { startDate, endDate },
+    dateRange: { startDate: start, endDate: end },
     refetch,
-    changeDateRange,
```
doneel, line 271: Unfortunately I believe this hook was used by Nav specifically so that they could change the date range via their own components.

What do you think we should do here...
1. Continue exposing a `changeDateRange` function that hooks into the global date?
2. Remove this, put migration instructions into a changelog & ping them?
3. Something else?

doneel (approved):

src/hooks/useProfitAndLoss/useProfitAndLoss.tsx
imkesin: The first option makes the most sense for now - I appreciate the catch (definitely bad on me).

PR #466, opened 2025-01-10
chore: lint scss imports

tomaszantas:
## Description

Replaces deprecated `@import` with `@forward`. It removes warning from the `npm run build`.

## Changes

- replace `@import` with `@forward` and remove `.scss` from imports in all scss files

## How this has been tested?

Before:

Screenshot 2025-01-10 at 14 45 29

After:

image


imkesin (approved):
This is great!

(I had no idea what the next move would be here)

PR #468, opened 2025-01-10
fix: `toDataProperties` needs greater typesafety to warn callers

imkesin:
## Description

This is my mistake for not making the function even more type-safe at the start.

I want callers to only use `kebab-case` for data properties... though this is a bit harder to enforce.


src/components/BankTransactionsTable/BankTransactionsTable.tsx
```diff
@@ -57,14 +57,14 @@ export const BankTransactionsTable = ({
...
+      && bankTransactions?.some(
         transaction => transaction.document_ids?.length > 0,
-      )) ??
-    false
+      ))
+      ?? false
 
   const showReceiptDataProperties = useMemo(
-    () => toDataProperties({ showReceiptUploadColumn: showReceiptColumn }),
+    () => toDataProperties({ 'show-receipt-upload-column': showReceiptColumn }),
```
imkesin, line 67: Notable change of input

tomaszantas (approved):
LGTM!

PR #473, opened 2025-01-14
revision: move revision changes to account balance inputs

tomaszantas:
## Description

Move selected changes from #463.

## Changes

Moved all changes not related to opening balance and selected changes for opening balance.
I pointed additional suggestion that we discussed on call and where not in the original revision.

## Blockers

None

## How this has been tested?

Regular usage on the LinkedAccounts view:

https://github.com/user-attachments/assets/fef415e5-316a-4980-b0b6-2f33b455c7ee

- shows anim on success, refetches external accounts and automatically closes the modal

On errors:

https://github.com/user-attachments/assets/26cdd1fc-e4d5-40aa-ad6e-bda83f239239

- validates empty inputs
- shows error when API request fails

### Extended case

https://github.com/user-attachments/assets/51c7e94c-6694-4c64-9252-d9c08647dc24

- shows success animation and hides successful forms
- show errors for invalid forms
- modal is automatically closed when all saved with success, and external accounts refetched
- accounts are automatically refetch on modal close when at least one form was saved with success


src/components/LinkedAccounts/OpeningBalanceModal/useUpdateOpeningBalanceAndDate.ts
```diff
@@ -50,84 +108,47 @@ export function useUpdateOpeningBalanceAndDate(
...
+            const errors = validate({ accountId, openingDate, openingBalance })
+            if (errors.length > 0) {
+              return Promise.reject(
+                new Error('Invalid data', {
+                  cause: {
+                    type: 'validation',
+                    errors,
+                    accountId,
+                  },
+                }),
```
tomaszantas, line 126: I'm extending the error content so it's easier to distinct validation and failed API calls

src/components/LinkedAccounts/OpeningBalanceModal/useUpdateOpeningBalanceAndDate.ts
```diff
@@ -50,84 +108,47 @@ export function useUpdateOpeningBalanceAndDate(
...
-        }
-        catch {
-          newErrors[item.account.id] = ['API_ERROR']
-          return
-        }
+      .then((results) => {
+        const resultsWithIds: OpeningBalanceAPIResponseResult[] = results.map((r, i) => ({
+          ...r,
+          accountId: data[i].accountId,
+        }))
```
tomaszantas, line 145: `Promise.allSettled` preserves the order of requests and my first thought was about using index. But on subsequent form submits, we don't sent again forms already saved, so number of items pass to `useSWRMutation` may not match number of records in the component state. Therefore matching by index won't work and `accountId` is needed.

src/components/LinkedAccounts/AccountFormBox/accountFormBox.scss
```diff
@@ -1,125 +1,122 @@
...
+  border-radius: var(--border-radius-xs);
+  border: 1px solid var(--color-base-100);
+  container-type: inline-size;
+  transition: all 0.25s ease-in-out;
+  transition-delay: 0.8s;
+  max-height: 280px;
+
+  &[data-confirmed='true'] {
+    background-color: var(--color-base-50);
+  }
```
imkesin, line 16: Note:

Also safe to just do `data-confirmed`.

I designed our data properties utility to never produce `data-x=false` - I strongly dislike when libraries do this instead of dropping the property.

src/components/LinkedAccounts/OpeningBalanceModal/useUpdateOpeningBalanceAndDate.ts
imkesin: Makes a ton of sense to me.

imkesin (approved):
Let's merge this into the base branch, then merge the base into `main`.

PR #474, opened 2025-01-16
chore: lint clean up 1

tomaszantas:
## Description

Small batch of eslint and stylelint fixes on selected folders:
- api
- config
- contexts
- models
- providers
- utils
- views

Stats:
- Stylelint -> was: 1163 | after: 1162 | to autofix: 1033
- Eslint -> was: 1570 | after: 1426 | to autofix: 1114

## To discuss

1. [?]  SCSS - move `*.scss`files to the component folders, except files like `variables.scss`, `utils.scss` etc.
2. [?] Use `type` over `interface`
3. [?] Do not nest single `*.tsx` files into folders (unless additional files are used), and don't use `index.tsx`

For example:

```
CORRECT:

-- contexts
    |-- BalanceSheetContext.tsx

------

CORRECT:

-- contexts
    |-- BalanceSheetContext
          |-- BalanceSheetContext.tsx
          |-- utils.ts

------

WRONG:

-- contexts
    |-- BalanceSheetContext
          |-- BalanceSheetContext.tsx
          |-- index.tsx

------
WRONG:

-- contexts
    |-- BalanceSheetContext
          |-- BalanceSheetContext.tsx

------
WRONG:

-- contexts
    |-- BalanceSheetContext
          |-- index.tsx

------


```


src/api/layer/authenticated_http.ts
```diff
@@ -1,12 +1,21 @@
...
 } as const
 
+type APIResponseError = {
+  description?: string
+  type?: string
+}
+
+type FailedAPIResponse = {
+  errors?: APIResponseError[]
+}
```
tomaszantas, line 17: Main change - improve types of API responses - because it comes from outside, all props are optional

src/providers/Environment/environmentConfigs.ts
```diff
@@ -1,5 +1,4 @@
-const ENVIRONMENTS = [ 'production', 'sandbox', 'staging', 'internalStaging' ] as const
-export type Environment = typeof ENVIRONMENTS[number]
+export type Environment = 'production' | 'sandbox' | 'staging' | 'internalStaging'
```
tomaszantas, line 1: Does it make sense to drop `const ENVIRONMENTS` entirely and move all to `Environment`?

imkesin: This change is fine for now.

If we ever want to reference the string literals at runtime, then we might go back to the approach of a const tuple.

src/views/AccountingOverview/internal/TransactionsToReview.tsx
```diff
@@ -46,18 +46,20 @@ export function TransactionsToReview({
 
   useEffect(() => {
     checkTransactionsToReview()
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
   useEffect(() => {
     checkTransactionsToReview()
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [dateRange, loaded])
```
imkesin, line 55: This is a massive code-smell... something to not be repeated in the future.

imkesin (approved):

PR #475, opened 2025-01-17
draft: establish a clean `tsconfig.json`

imkesin:
## Task

Use `npm run typecheck` or `npx tsc --noEmit` to figure out where there are type errors in the repo.

A significant chunk of them are related to `noUncheckedIndexAccess`.

I encourage you to use type guards when possible:
```
type ArrayWithAtLeastTwo<T> = readonly [T, T, ...T[]]

export function isArrayWithAtLeastTwo<T>(list: ReadonlyArray<T>): list is ArrayWithAtLeastTwo<T> {
  return list.length >= 2
}
```

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,106 +1,34 @@
 {
   "compilerOptions": {
-    /* Visit https://aka.ms/tsconfig to read more about this file */
-    /* Projects */
-    // "incremental": true,                              /* Save .tsbuildinfo files to allow for incremental compilation of projects. */
-    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -1,106 +1,34 @@
...
+   "esModuleInterop": true,
+   "skipLibCheck": true,
+   "target": "es2022",
+   "allowJs": true,
+   "resolveJsonModule": true,
+   "moduleDetection": "force",
+
+   // Uncomment and resolve ASAP
+   // "isolatedModules": true,
+   // "verbatimModuleSyntax": true,
```
imkesin, line 13: These two need to be addressed in a follow-on PR.

They mostly related to the usage of `import type` syntax.

imkesin:
This has been superceded by #625

PR #476, opened 2025-01-17
build: remove "engines" from `package.json`

imkesin:
## Description

This was a mistake by me.

The `"engines"` field in `package.json` conveys information about who should be running the package, which is mostly relevant for server code.

We should just maintain our `.nvmrc` file for development purposes.


src/hooks/useQuickbooks/useQuickbooks.ts
```diff
@@ -27,7 +27,7 @@ export const useQuickbooks: UseQuickbooks = () => {
   // Poll the server to determine when the Quickbooks sync is complete
   useEffect(() => {
     if (isSyncingFromQuickbooks && syncStatusIntervalRef.current === null) {
-      const interval = setInterval(() => fetchIsSyncingFromQuickbooks(), 2000)
+      const interval = window.setInterval(() => fetchIsSyncingFromQuickbooks(), 2000)
```
imkesin, line 30: This is an unrelated type error that needed to be fixed. When using browser globals, it is preferred to reference the `window` object.

doneel (approved):

PR #482, opened 2025-01-22
[LAY-1153] fix: load all categories for manual journal entries

imkesin:
## Description

**Linear**: LAY-1153

We are now loading journal entries from the categories endpoint.

### (Unrelated) Known Issues

If a user doesn't touch the date selector at all, no date has been set.

### DIFF
--- a/src/components/JournalForm/JournalFormEntryLines.tsx
+++ b/src/components/JournalForm/JournalFormEntryLines.tsx
@@ -1,5 +1,4 @@
-import { useContext } from 'react'
-import { ChartOfAccountsContext } from '../../contexts/ChartOfAccountsContext'
+import { useContext, useMemo } from 'react'
 import { JournalContext } from '../../contexts/JournalContext'
 import Trash from '../../icons/Trash'
 import { Direction, JournalEntryLineItem } from '../../types'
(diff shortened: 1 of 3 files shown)


src/components/JournalForm/JournalFormEntryLines.tsx
```diff
@@ -12,10 +11,21 @@ import {
...
+
+type WithSubCategories = { subCategories?: Array<WithSubCategories> }
+
+function recursiveFlattenCategories<T extends WithSubCategories>(accounts: Array<T>): Array<T> {
+  const flattenedResult = accounts.flatMap(a => [
+    a,
+    recursiveFlattenCategories((a.subCategories ?? [])),
+  ]).flat()
+
+  return flattenedResult as Array<T>
```
imkesin, line 27: I don't know how to make typescript properly infer the type here, so I need an assertion. I try to always very clearly separate out assertions.

src/components/JournalForm/JournalFormEntryLines.tsx
```diff
@@ -32,15 +42,59 @@ export const JournalFormEntryLines = ({
...
+            value: relevantCategory.id,
+            display_name: relevantCategory.display_name,
+          },
+          sub_accounts: [],
+          balance: 0,
+          normality: Direction.DEBIT,
+        },
+      ],
+    )
+  }
```
imkesin, line 97: This is absolutely awful.

The callback requires passing additional state just to make sense of the action. Since it doesn't clearly "pick" the fields that matter, I have to pass strange values to make typescript happy.

doneel: Yeah I see the issue - this feels like the right balance to me of "it's imperfect but gets the job done right now".

src/hooks/categories/useAllCategories.ts
```diff
@@ -0,0 +1,44 @@
...
+  apiUrl?: string
+  businessId: string
+}) {
+  if (accessToken && apiUrl) {
+    return {
+      accessToken,
+      apiUrl,
+      businessId,
+      tags: ['#categories'],
+    }
```
doneel, line 21: I can see this isn't the same key we use for the other `/categories` call (without `mode: ALL`), but for future proofing is it worth including something about the mode to differentiate this from the swr cache for the other call?

doneel (approved):

src/components/JournalForm/JournalFormEntryLines.tsx
```diff
@@ -32,15 +42,59 @@ export const JournalFormEntryLines = ({
...
+    const parentOptions = flattenedCategories
+      .sort((a, b) => (a.display_name.localeCompare(b.display_name)))
+      .map(({ display_name, id }) => {
+        return {
+          label: display_name,
+          value: id,
+        }
+      })
+
+    return { flattenedCategories, parentOptions }
```
tomaszantas, line 65: NIT: I can imagine that flattening categories could be a common thing to do across the application, so it could be moved to `src/utils/categories.ts`

tomaszantas (approved):

src/hooks/categories/useAllCategories.ts
imkesin: This is a great call-out. I'll make a fix.

PR #484, opened 2025-01-22
fix: alpha release fixes

imkesin:
## Description

**Linear**: LAY-1154 and LAY-1158

- Unified date range will not allow end dates beyond today
- Explicitly set a default date in the journal form

## Additional Commits

- Fixes issue where accounts without an ID - only a `STABLE_NAME` - are not handled properly in the dropdown.

### DIFF
--- a/src/components/JournalForm/JournalFormEntryLines.tsx
+++ b/src/components/JournalForm/JournalFormEntryLines.tsx
@@ -15,16 +15,19 @@ import { InputWithBadge, InputGroup, Select } from '../Input'
 import { JournalConfig } from '../Journal/Journal'
 import { Text, TextSize } from '../Typography'
 import { useAllCategories } from '../../hooks/categories/useAllCategories'
+import { safeAssertUnreachable } from '../../utils/switch/safeAssertUnreachable'
 
-type WithSubCategories = { subCategories?: Array<WithSubCategories> }
+type WithSubCategories = { subCategories: ReadonlyArray<WithSubCategories> | null }
 
-function recursiveFlattenCategories<T extends WithSubCategories>(accounts: Array<T>): Array<T> {
+function recursiveFlattenCategories<T extends WithSubCategories>(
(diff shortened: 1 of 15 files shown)


src/components/JournalForm/JournalFormEntryLines.tsx
```diff
@@ -53,12 +56,27 @@ export const JournalFormEntryLines = ({
...
+              label: account.display_name,
+              value: account.stable_name,
+            }
+          case 'ExclusionNested':
+            return {
+              label: account.display_name,
+              value: account.id,
+            }
+          default:
+            safeAssertUnreachable(account, 'Unexpected account type')
```
doneel, line 79: For my curiosity, why do it via explicit case handling vs. setting the `value` field to a polymorphic object? I'm imagining there could be an advantage in not having to do another case statement to *re*-construct the structure below where you set `const baseFields`

src/utils/journal.ts
```diff
@@ -5,19 +5,20 @@ import { JournalEntry, JournalEntryLineItem } from '../types/journal'
...
     }
   }
 
+  if (journalLineItem.account_identifier.id) {
+    return {
+      type: 'AccountId',
+      id: journalLineItem.account_identifier.id,
+    }
+  }
+
```
doneel, line 21: nice this is what I expect we have to do at the end of the flow

imkesin: Yup - I think we intentionally want the "flipped" version to help us prioritize `STABLE_NAME` usage.

src/components/JournalForm/JournalFormEntryLines.tsx
imkesin: I didn't even think to do something like this:

```
class OptionClass {

    get label() {
        return display_name
    }

    get value() {
        return stable_name
    }
}
```

I agree that the double-switch is a code smell.

PR #486, opened 2025-01-23
feat: bills and vendors types and utils

tomaszantas:
## Description

This PR brings smaller changes from a large Bills PR.

## Changes

- Bills and Vendors types
- Bills and Vendors helper functions (utils)

### DIFF
--- a/src/types/bills.ts
+++ b/src/types/bills.ts
@@ -8,81 +8,81 @@ export type BillStatusFilter = 'PAID' | 'UNPAID'
 export const UnpaidStatuses = ['SENT', 'PARTIALLY_PAID']
 export const PaidStatuses = ['PAID']
 
-export const BillTerms = [{
-  id: 'due_on_receipt',
+type BillTerm = 'DUE_ON_RECEIPT' | 'NET_10' | 'NET_15' | 'NET_30' | 'NET_60'
+
+export const BillTerms: {
+  id: BillTerm
+  label: string
+}[] = [{
+  id: 'DUE_ON_RECEIPT',
   label: 'Due on receipt',
 }, {
-  id: 'net_10',
+  id: 'NET_10',
   label: 'Net 10',
 }, {
-  id: 'net_15',
+  id: 'NET_15',
   label: 'Net 15',
 }, {
-  id: 'net_30',
+  id: 'NET_30',
   label: 'Net 30',
 }, {
-  id: 'net_60',
+  id: 'NET_60',
   label: 'Net 60',
 }]
 
-export type PaymentAllocationMethod =
-  | 'CASH'
-  | 'CHECK'
-  | 'CREDIT_CARD'
-  | 'ACH'
-  | 'REDEEMED_PREPAYMENT'
-  | 'OTHER'
-
 export type Bill = {
-  id: string
-  type: 'Bill'
-  business_id: string
-  external_id: string
-  status: BillStatus
-  received_at: string
-  due_at: string
-  paid_at?: string
-  voided_at?: string
+  additional_sales_taxes: SalesTax[]
+  additional_sales_taxes_total: number
   bill_number: string
-  line_items: BillLineItem[]
-  subtotal: number
-  total_amount: number
-  outstanding_balance: number
-  payment_allocations: BillPayment[]
+  business_id: string
+  due_at: string
+  external_id: string
+  id: string
   imported_at: string
-  updated_at: string
+  line_items: BillLineItem[]
+  outstanding_balance: number
+  paid_at?: string
+  payment_allocations: BillPaymentAllocation[]
+  received_at: string
+  status: BillStatus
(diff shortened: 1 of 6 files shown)


src/types/bills.ts
```diff
@@ -0,0 +1,88 @@
...
+  received_at: string
+  due_at: string
+  paid_at?: string
+  voided_at?: string
+  bill_number: string
+  line_items: BillLineItem[]
+  subtotal: number
+  total_amount: number
+  outstanding_balance: number
+  payment_allocations: BillPayment[]
```
AndrewAcomb, line 51: Note- we will be using a separate type for BillPaymentAllocation. I have a PR out with some updates to our docs for AP - https://github.com/Layer-Fi/api-documentation/pull/105

AndrewAcomb: We will also have tax line items on bills and bill line items (see ^ pr)

AndrewAcomb (changes requested):
Could you add the fields added in this PR?

https://github.com/Layer-Fi/api-documentation/pull/105

Namely tax line items and bill payment allocations

src/types/bills.ts
imkesin, line 1: (Minor - no need to change)

I think we should stop doing the `types/*` files altogether. I'd prefer we co-locate these with data-loading. The specific advantage would be applying transformations right at the top-level (e.g. ISO string => date).

src/types/bills.ts
```diff
@@ -0,0 +1,88 @@
+import { TransactionTag } from './tags'
+import { Vendor } from './vendors'
+
+export type BillStatus = 'SENT' | 'PARTIALLY_PAID' | 'PAID' | 'VOIDED'
+
+export type BillStatusFilter = 'PAID' | 'UNPAID'
+
+export const UnpaidStatuses = ['SENT', 'PARTIALLY_PAID']
+export const PaidStatuses = ['PAID']
```
imkesin, line 9: The declaration ordering doesn't make sense here.

```
const UNPAID_STATUS_MAP = {
  SENT: 'SENT',
  PARTIALLY_PAID: 'PARTIALLY_PAID',
  EXTRA_FIELD: 'EXTRA_FIELD',
} as const
type UnpaidBillStatus = typeof UNPAID_STATUS_MAP[keyof typeof UNPAID_STATUS_MAP]
const UNPAID_BILL_STATUSES = [
  UNPAID_STATUS_MAP.SENT,
  UNPAID_STATUS_MAP.PARTIALLY_PAID,
] as const

const PAID_STATUS_MAP = {
  PAID: 'PAID',
} as const
type PaidBillStatus = typeof PAID_STATUS_MAP[keyof typeof PAID_STATUS_MAP]
const PAID_BILL_STATUSES = [PAID_STATUS_MAP.PAID] as const
```

This setup would have the types flow from their concrete declaration into types.

src/types/bills.ts
```diff
@@ -0,0 +1,88 @@
...
+  received_at: string
+  status: BillStatus
+  subtotal: number
+  terms: BillTerm
+  total_amount: number
+  transaction_tags: TransactionTag[]
+  type: 'Bill'
+  updated_at: string
+  voided_at?: string
+  vendor?: Vendor
```
imkesin, line 55: Nit:

I think we should be careful with our use of optional properties. I've seen the API mostly returning `null` - that would be more representative of the real state.

src/utils/bills.ts
```diff
@@ -0,0 +1,17 @@
+import { BillStatus, PaidStatuses, UnpaidStatuses } from '../types/bills'
+
+export const isBillPaid = (status?: BillStatus) => {
+  if (!status) {
+    return false
+  }
+
+  return PaidStatuses.includes(status)
+}
```
imkesin, line 9:
```
export function alt_isBillPaid<T extends Pick<Bill, 'status'>>(
  bill: T,
): bill is typeof bill & { status: 'PAID' } {
  return PaidStatuses.includes(bill.status)
}
```

These types of functions should narrow when possible.

src/types/bills.ts
tomaszantas: Do you think about having types in `hooks/*` and/or `api/*` folders (roughly speaking)?
Both approaches works for me. Having it collocated with data-loading has obvious advantage. But having types in the `types/*` makes it easier to identify main and commonly used (data) types.

src/utils/bills.ts
tomaszantas: I changed these a bit due to changes in types, but suggested code above didn't work:

```
Argument of type 'BillStatus' is not assignable to parameter of type '"SENT" | "PARTIALLY_PAID"'.
  Type '"PAID"' is not assignable to type '"SENT" | "PARTIALLY_PAID"'.
```

imkesin: (We discussed a revision to this in a call on Monday, Jan 27th)

PR #487, opened 2025-01-23
feat: extending components - for bills

tomaszantas:
## Description

This PR brings a chunk of smaller changes to existing components that will be then used with the Bills.
This is a part of efforts aiming to minify large Bills PR.

## Changes

- allow to control button variant in `SubmitButton`
- make DatePicker's `maxDate` optional
- allow to use `HeaderCol` without paddings
- add `StaticValue` - it can be use as a replacement to the (disabled) inputs, ie. on summary view after submitting the form when we want to keep similar layout, but showing disabled input doesn't look right. `StaticValue` has similar sizings as `Input`
- add `nowrap` prop to `TableCell` so the cell content is not wrapped
- fix `Toggle` state management
- add `ellipsis` prop to `Text` so it sets `text-overflow: ellipsis`
- add `status` prop to `Text` so we can easily set basic colors like error, success, warning
- clean up `inputs.scss` and `tables.scss` files

NOTE: [Low-Priority] Some components touched in this PR, like `Text`, requires further refactors in new PRs, ie. switching from css classes to data-attributes.

### DIFF
--- a/src/components/Header/HeaderCol.tsx
+++ b/src/components/Header/HeaderCol.tsx
@@ -1,14 +1,23 @@
-import { CSSProperties, ReactNode } from 'react'
+import { CSSProperties, ReactNode, useMemo } from 'react'
 import classNames from 'classnames'
+import { toDataProperties } from '../../utils/styleUtils/toDataProperties'
 
 interface HeaderColProps {
   className?: string
   style?: CSSProperties
+  noPadding?: boolean
   children: ReactNode
 }
 
-export const HeaderCol = ({ className, children, style }: HeaderColProps) => (
-  <div className={classNames('Layer__header__col', className)} style={style}>
-    {children}
-  </div>
-)
+export const HeaderCol = ({ className, children, style, noPadding = false }: HeaderColProps) => {
+  const dataProperties = useMemo(
+    () => toDataProperties({ 'no-padding': noPadding }),
+    [noPadding],
+  )
+
+  return (
+    <div {...dataProperties} className={classNames('Layer__header__col', className)} style={style}>
+      {children}
+    </div>
+  )
+}
--- a/src/components/Header/HeaderCol.tsx
+++ b/src/components/Header/HeaderCol.tsx
@@ -1,4 +1,4 @@
-import { CSSProperties, ReactNode, useMemo } from 'react'
+import { CSSProperties, ReactNode } from 'react'
 import classNames from 'classnames'
 import { toDataProperties } from '../../utils/styleUtils/toDataProperties'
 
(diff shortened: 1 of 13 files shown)


src/components/Header/HeaderCol.tsx
```diff
@@ -1,14 +1,23 @@
...
-export const HeaderCol = ({ className, children, style }: HeaderColProps) => (
-  <div className={classNames('Layer__header__col', className)} style={style}>
-    {children}
-  </div>
-)
+export const HeaderCol = ({ className, children, style, noPadding = false }: HeaderColProps) => {
+  const dataProperties = useMemo(
+    () => toDataProperties({ 'no-padding': noPadding }),
+    [noPadding],
+  )
```
AndrewAcomb, line 16: toDataProperties does not strike me as particularly expensive - why do we need to memoize here?

AndrewAcomb (approved):

src/components/Header/HeaderCol.tsx
tomaszantas: You are right! thanks for this catch

src/styles/header.scss
```diff
@@ -33,10 +33,15 @@
   padding: var(--spacing-md);
   min-height: 68px;
   box-sizing: border-box;
+
+  &[data-no-padding='true'] {
+    padding-left: 0;
+    padding-right: 0;
+  }
```
imkesin, line 40: (No need to change)

I would've implemented this as `data-pi=none`

src/styles/table.scss
```diff
@@ -1,5 +1,6 @@
 .Layer__table {
   width: 100%;
+  border-collapse: separate;
```
imkesin, line 3: Very cool - I didn't know about this property

src/styles/typography.scss
```diff
@@ -57,12 +56,31 @@
 .Layer__text--bold,
 .Layer__text.Layer__text--bold {
   font-weight: var(--font-weight-bold);
+  font-variation-settings: 'wght' 540;
```
imkesin, line 59: Nit:

I don't want us to use this in any spot - this type of property can't be overridden (breaks the CSS mental model).

src/components/Toggle/Toggle.tsx
```diff
@@ -55,21 +55,23 @@ export const Toggle = ({
   const [currentWidth, setCurrentWidth] = useState(0)
   const [thumbPos, setThumbPos] = useState({ left: 0, width: 0 })
   const [initialized, setInitialized] = useState(false)
+  const [activeOption, setActiveOption] = useState(selected || options[0].value)
```
imkesin, line 58: Nit:

Are the toggle options guaranteed to exist? Can they ever be dynamically loaded (and the array of options will be empty)?

src/components/Toggle/Toggle.tsx
imkesin, line 1: If this is going to become a component we change often, it might be worth cutting over to a [dedicated Tabs component](https://react-spectrum.adobe.com/react-aria/Tabs.html).

imkesin (approved):

src/styles/typography.scss
tomaszantas: You are right! Switching to `var`

PR #490, opened 2025-02-04
feat: pnl compare - disable compare when year picker

tomaszantas:
## Description

In Profit & Loss report, disable the compare mode when picker mode is different than `monthPicker` or `yearPicker`.

## How this has been tested?

https://github.com/user-attachments/assets/acb47eff-eb9a-479d-bbdb-3d3659a3c89a

<br />

### What if day range mode enabled...

When different picker mode enable in Profit and Loss DatePicker, then the compare mode is disabled and table is goes into regular structure.

https://github.com/user-attachments/assets/35d761c7-f786-471e-b9be-8f3d0edc1529


src/components/ProfitAndLossCompareOptions/ProfitAndLossCompareOptions.tsx
imkesin, line 1: We need to remove all `useEffect` calls from this file.

It should be possible to derive the state here.

src/components/ProfitAndLossCompareOptions/ProfitAndLossCompareOptions.tsx
```diff
@@ -33,70 +34,79 @@ const selectStyles = {
...
 
-  const { dateRange } = useContext(ProfitAndLoss.Context)
+  const { rangeDisplayMode } = useGlobalDateRange()
+
+  const [periods, setPeriods] = useState<number>(
+    comparePeriods !== 0 ? comparePeriods : 1,
+  )
+  const [toggle, setToggle] = useState<TagComparisonOption[]>(
+    compareOptions?.length > 0 ? compareOptions : [defaultOption],
+  )
```
imkesin, line 63: Nit:

This is a strange name for the selected tag comparison options.

src/hooks/useProfitAndLossComparison/useProfitAndLossComparison.tsx
imkesin, line 1: Very good clean up in this file.

The next step will be to completely get rid of the contexts.

src/utils/profitAndLossComparisonUtils.ts
```diff
@@ -1,7 +1,19 @@
...
 
-export const generatComparisonMonths = (
+export const generateComparisonPeriods = (
+  startDate: Date, numberOfPeriods: number, rangeDisplayMode: DateRangePickerMode,
+) => {
+  switch (rangeDisplayMode) {
+    case 'yearPicker':
+      return generatComparisonYears(startDate, numberOfPeriods)
+    default:
+      return generatComparisonMonths(startDate, numberOfPeriods)
```
imkesin, line 12: Nit: spelling

src/hooks/useProfitAndLossComparison/useProfitAndLossComparison.tsx
```diff
@@ -20,179 +17,83 @@ type Props = {
...
 
 export function useProfitAndLossComparison({
   reportingBasis,
 }: Props) {
-  const lastQuery =
-    useRef<
-      Promise<{ data?: ProfitAndLossComparison | undefined, error?: unknown }>
-    >()
-
   const [compareMode, setCompareMode] = useState(false)
```
imkesin, line 51: I'm confused by this.

It's always a comparison PnL (i.e. - it hits the same endpoint), but one mode just limits a single dimension.

tomaszantas: I changed the name and switched to `useMemo` since it's actually evaluated based on the combination of periods and options values.

src/components/ProfitAndLossCompareOptions/ProfitAndLossCompareOptions.tsx
```diff
@@ -34,82 +14,31 @@ const selectStyles = {
...
-    else if (JSON.stringify(toggle) !== JSON.stringify(compareOptions) && setCompareOptions) {
-      setCompareOptions(toggle)
-    }
-  }, [toggle, compareOptions, setCompareOptions, defaultOption])
-
   const timeComparisonOptions = [
     { value: 1, label: rangeDisplayMode === 'monthPicker' ? 'This month' : 'This year' },
     { value: 2, label: rangeDisplayMode === 'monthPicker' ? 'Last 2 months' : 'Last 2 years' },
     { value: 3, label: rangeDisplayMode === 'monthPicker' ? 'Last 3 months' : 'Last 3 years' },
   ]
```
imkesin, line 39: I think this still doesn't cover all of the possible cases.

If we select a custom date range, then I think we should default to `This period` (i.e. we need an additional case to check for the `=== 'yearPicker'`.

imkesin: This screenshot shows a state that I didn't expect we could enter:

Screenshot 2025-02-13 at 1 45 27 PM

The presence of the tag filter doesn't make much sense in this view.

src/components/ProfitAndLossCompareOptions/index.tsx
```diff
@@ -1,4 +1 @@
-export {
-  ProfitAndLossCompareOptions,
-  ProfitAndLossCompareOptionsProps,
-} from './ProfitAndLossCompareOptions'
+export { ProfitAndLossCompareOptions } from './ProfitAndLossCompareOptions'
```
imkesin, line 1: Nit:

If we can, let's just kill a barrel file entirely.

src/hooks/useProfitAndLossComparison/useProfitAndLossComparison.tsx
```diff
@@ -47,29 +55,63 @@ function buildKey({
...
-  const [compareOptions, setCompareOptions] = useState<TagComparisonOption[]>(
-    [],
+  const [comparePeriods, setComparePeriods] = useState(comparisonConfig?.defaultPeriods ?? 1)
+  const [selectedCompareOptions, setSelectedCompareOptionsState] = useState<TagComparisonOption[]>(
+    comparisonConfig?.defaultTagFilter ? [comparisonConfig?.defaultTagFilter] : [],
   )
   const { rangeDisplayMode, start, end } = useGlobalDateRange()
   const dateRange = { startDate: start, endDate: end }
 
+  const isCompareDisabled = !ALLOWED_COMPARE_MODES.includes(rangeDisplayMode)
```
imkesin, line 67: I think this is not the correct assertion.

If we have tags selected, then we should still be presenting a comparison PnL.

src/components/ProfitAndLossCompareOptions/ProfitAndLossCompareOptions.tsx
tomaszantas: Good suggestion with showing `This period`, thanks! I changed it into `switch-case` with cases for `month`, `year`, and default fallback into `period`.

State from the screenshot comes from global date after switching dates in Balance sheet. I agree that's not the expected behavior here, and I'm going to resolve this issue with #497

imkesin (approved):
I added a small commit myself.

PR #493, opened 2025-02-08
fix: we need to explicitly set the `color` property in many locations

imkesin:
## Description

One of our customers relies on a CSS-reset that makes both the `color` and `background-color` transparent by default. This exposes that we have many areas where we are relying too heavily on a cascade of color.

This first block of changes is intended to color getting the text on the summary cards to appear correctly. This should have (minimal) impact + zero logical changes.


src/components/ui/Typography/heading.scss
```diff
@@ -1,13 +1,22 @@
...
+    font-weight: var(--font-weight-bold);
+  }
+
   &[data-size="sm"] {
     font-size: var(--text-heading-sm);
   }
 
+  &[data-size="lg"] {
+    font-size: var(--text-heading-secondary);
+  }
```
imkesin, line 14: The name `--text-heading-secondary` makes no sense.

We need to find a way to standardize the spectrum of sizes.

AndrewAcomb (approved):

tomaszantas (approved):

PR #498, opened 2025-02-13
feat: business details form

tomaszantas:
## Description

Part of Onboarding: This PR moves Business details form from old PR using Tanstack form.

This PR builds only the Form without handling API communication.

To discuss:
- full name OR first name + last name
- us state - 2 letters code
- assign default role

## Changes

- install Tanstack form package
- add `BusinessForm`
- change `Business (type)` fields: `us_state: string -> USState` and `entity_type: string -> EntityType` 
- add `BusinessTypeSelect` allowing to select `EntityType`
- add `USStateSelect` allowing to select `USState`
- add `PhoneInput` - handles content formatting
- add `FormSection` - creates section with a given title and puts inputs into the column 

## How this has been tested?

Screenshot 2025-02-26 at 15 03 03

Screenshot 2025-02-26 at 15 03 29


src/types/location.ts
```diff
@@ -0,0 +1,56 @@
+export const US_STATES = [
```
imkesin, line 1: This needs to switch to map of from the ansi abbreviations ([link to Wikipedia](https://en.wikipedia.org/wiki/List_of_U.S._state_and_territory_abbreviations)) to the full name of the location.

Something like...
```
{
  "AL": {
     name: "Alabama"
   },
}
```

src/components/BusinessForm/useBusinessForm.ts
```diff
@@ -0,0 +1,45 @@
...
+  legal_name?: string
+  dba?: string
+  entity_type?: EntityType
+  us_state?: USState
+  tin?: string
+}
+
+export const useBusinessForm = () => {
+  const { business } = useLayerContext()
+  const form = useForm<
```
imkesin, line 21: The form can be incredibly difficult to get working correctly with type inference (I've tried).

This limitation means that it makes more sense to inline.

src/components/Input/PhoneInput.tsx
```diff
@@ -0,0 +1,47 @@
...
+  value?: string
+  onChange: (value?: string) => void
+}
+
+export const PhoneInput = ({
+  value,
+  onChange,
+  placeholder = 'Phone number',
+  ...props
+}: PhoneInputProps) => {
```
imkesin, line 14: This type of component requires [libphonenumber-js](https://www.npmjs.com/package/libphonenumber-js) for proper handling.

imkesin (commented):
I want to pause this until we can more clearly wire it up with endpoints.

src/types/location.ts
tomaszantas: [API](https://docs.layerfi.com/api-reference/v1/update-business#body-us-state) uses full names, so not sure if using ISO or ANSI code is necessary in this case.

src/components/Input/PhoneInput.tsx
tomaszantas: I have hesitation on adding another dependency if same thing can be easily achieve with own code (in overall).
If we support US phone numbers only, then current implementation makes more sense to me. However, if we are going to support other phone numbers, then having battle-tested library is definitely better choice

package-lock.json
```diff
@@ -10,8 +10,10 @@
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@floating-ui/react": "^0.26.8",
+        "@tanstack/react-form": "^0.43.0",
```
imkesin, line 13: I think this will go to `1.0.0` after a rebase.

tomaszantas: That's right

src/hooks/business/useUpdateBusiness.ts
```diff
@@ -0,0 +1,74 @@
...
+import { useAuth } from '../useAuth'
+import { useLayerContext } from '../../contexts/LayerContext'
+import {
+  updateBusiness,
+  type UpdateBusinessBody,
+} from '../../api/layer/business'
+import { useSWRConfig } from 'swr'
+import { withSWRKeyTags } from '../../utils/swr/withSWRKeyTags'
+
+export const BUSINESS_TAG_KEY = 'business'
```
imkesin, line 11: (No change for now)

It would be sweet if we could find a way to extract `Layer.getBusiness` so that we could nicely represent the relationships between the hooks.

imkesin (approved):

PR #499, opened 2025-02-19
refactor: pagination

tomaszantas:
## Description

Refactoring Pagination component.

Pagination was wrapped into additional containers with the same styling in many components (ie. BankTransactions, Journal, etc.). This PR moves that container into `<Pagination />` component.

This change is also implemented in the Bills PR, but that change touches many files and most changes were produced by linters. So to keep Bills PR small as possible and easier to review, I'm moving this refactor to this PR.

## Changes

- wrap Pagination into the container giving proper spacing and positioning across the application
- remove styles and refactor related components

## How this has been tested?

image

image

### DIFF
--- a/src/components/BankTransactions/BankTransactions.tsx
+++ b/src/components/BankTransactions/BankTransactions.tsx
@@ -34,7 +34,7 @@ export type BankTransactionsMode = 'bookkeeping-client' | 'self-serve'
 
 export const categorizationEnabled: (
   mode: BankTransactionsMode,
-) => boolean = mode => {
+) => boolean = (mode) => {
   if (mode === 'bookkeeping-client') {
     return false
   }
@@ -100,7 +100,7 @@ const BankTransactionsContent = ({
   hideHeader = false,
   stringOverrides,
 }: BankTransactionsProps) => {
-  const [ defaultDateRange ] = useState(() => ({
+  const [defaultDateRange] = useState(() => ({
     startDate: startOfMonth(new Date()),
     endDate: endOfMonth(new Date()),
   }))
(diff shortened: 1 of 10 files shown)


src/components/BankTransactions/BankTransactions.tsx
```diff
@@ -318,61 +324,67 @@ const BankTransactionsContent = ({
...
+            isValidating={isValidating}
+            error={error}
+            refetch={refetch}
+            editable={editable}
+          />
+        )
+        : null}
 
       {!monthlyView && (
-        <div className='Layer__bank-transactions__pagination'>
```
tomaszantas, line 366: Removing `<div className='Layer__bank-transactions__pagination'>` is actual change in this file. Other changes are lint fixes

src/styles/pagination.scss
imkesin, line 1: (No need to change)

This would be a good opportunity to co-locate this within the `/Pagination` folder.

src/styles/pagination.scss
```diff
@@ -1,11 +1,26 @@
...
+  justify-content: flex-end;
+  padding: var(--spacing-md);
+  position: sticky;
+  left: 0;
+
+  @container (min-width: 1400px) {
+    display: flex;
+    justify-content: flex-end;
+    padding: var(--spacing-md) var(--spacing-xl);
+  }
```
imkesin, line 12: I didn't know this was allowed... I should be using this more often instead of writing separate `@` blocks.

src/styles/pagination.scss
```diff
@@ -1,11 +1,26 @@
+.Layer__pagination-container {
+  display: flex;
+  justify-content: flex-end;
+  padding: var(--spacing-md);
+  position: sticky;
```
imkesin, line 5: (To be looked at)

I am slightly suspicious of `sticky`. Based on how we're using the container, I don't see any dramatic use-case for sticky.

imkesin (approved):

PR #500, opened 2025-02-20
feat: `toDefinedSearchParameters` should handle lists

imkesin:
## Description

Our function that builds query parameters needs to anticipate transforming arrays into `value=1&value=2`.

### DIFF
--- a/src/utils/request/toDefinedSearchParameters.ts
+++ b/src/utils/request/toDefinedSearchParameters.ts
@@ -7,31 +7,40 @@ function toSnakeCase(input: string) {
     .join('_')
 }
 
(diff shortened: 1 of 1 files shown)


tomaszantas (approved):
The only thought I have is about using `value=1&value=2` instead of `value[]=1&value[]=2`.

For instance:
* `value=1&value=2` is translated into array: `value=[1,2]`, same as `value[]=1&value[]=2`
* ...but `value=1` is just a string/number: `value=1`
* ...and `value[]=1` is single element array: `value=[1]`

Which way is better actually depends on how API works and both can work, so I'm leaving this judgment to you.

PR #357, opened 2024-10-03
feat: add bills view

ArkuVonSymfon:
## Description

Add Bills

To do in following up PRs:
- Category select in Bills Details should use API to get available options instead of hiding "REVENUE" categories on UI
- add create Bill

## How this has been tested?

https://github.com/user-attachments/assets/69a5e5b5-a678-4dec-a86a-8803a93c21eb

Screenshot 2025-02-24 at 16 43 20

Screenshot 2025-02-24 at 16 43 34

PR #504, opened 2025-02-25
feat: bills and vendors api, vendor select

tomaszantas:
## Description

Selected changes from the Bills PR #357 are moved here to make the original PR easier to review.

## Changes

- add API functions for Bills and Vendors
- add VendorSelect component
- update Bills types

## How this has been tested?

image

### DIFF
--- a/src/types/bills.ts
+++ b/src/types/bills.ts
@@ -56,7 +56,7 @@ export type Bill = {
   line_items: BillLineItem[]
   outstanding_balance: number
   paid_at?: string
-  payment_allocations: PaymentAllocation[]
+  payment_allocations: BillPaymentAllocation[]
   received_at: string
   status: BillStatus
   subtotal: number
@@ -73,7 +73,6 @@ export type BillLineItem = {
   account_identifier?: {
     type: string
     id: string
-    product_name?: string
   }
   bill_id: string
   description: string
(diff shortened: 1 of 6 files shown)


src/types/bills.ts
```diff
@@ -56,7 +56,7 @@ export type Bill = {
   line_items: BillLineItem[]
   outstanding_balance: number
   paid_at?: string
-  payment_allocations: BillPaymentAllocation[]
+  payment_allocations: PaymentAllocation[]
   received_at: string
```
AndrewAcomb, line 60: I would keep calling it BillPaymentAllocation because we will at some point be adding AR (invoices) which has an analogous concept of InvoicePaymentAllocation

src/types/bills.ts
```diff
@@ -65,25 +65,31 @@ export type Bill = {
...
 }
 
-type BillLineItem = {
-  account_identifier: string
+export type BillLineItem = {
+  account_identifier?: {
+    type: string
+    id: string
+    product_name?: string
+  }
```
AndrewAcomb, line 77: product_name should not be included in the account identifier object. Also I believe this should just be the same AccountIdentifier type we use in bank transactions (`AccountIdentifierPayloadObject`)

AndrewAcomb (approved):
a few minor changes (also is selectVendor empty intentionally?) but LGTM otherwise

PR #506, opened 2025-02-25
feat: useSWRLoadingStatus

tomaszantas:
## Description

The `useSWR` doesn't set `isLoading = true` at the very beginning until it runs actual query. It may happen for instance when access token request has not been completed yet. Then on UI, we may display for a moment empty state instead of loading state. The `useSWRLoading` hook replaces `useSWR`'s `isLoading` and sets it to true also before the first request is being sent.

### DIFF
--- a/src/hooks/useSWRLoadingStatus.ts
+++ b/src/hooks/useSWRLoadingStatus.ts
@@ -1,0 +1,44 @@
+import { LoadedStatus } from '../types/general'
+import { useEffect, useState } from 'react'
+
+type UseSWRLoadingStatusProps = {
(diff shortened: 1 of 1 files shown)


imkesin:
Can you point me to a real example of this behavior?

This sounds like it is coming from us not understanding the waterfall of `useAuth` loading into `useOtherHook` loading.

imkesin (changes requested):
I need to see an example / will make separate improvements to our use of SWR if needed.

tomaszantas:
You are right, this new hook is not necessary

PR #507, opened 2025-02-25
feat: modal default size, checkbox variants

tomaszantas:
## Description

- change the default modal size 
- add success and dark variations to the checkbox
- add large and small variations to the checkbox
- add optional tooltip to the checkbox (ie. can be used to explain why checkbox is disabled)

## How this has been tested?

Modal default size before:

Screenshot 2025-02-25 at 18 37 17

Modal default size after:

Screenshot 2025-02-25 at 18 44 20

Checkboxes:

image

Screenshot 2025-02-25 at 18 52 18

Screenshot 2025-02-25 at 18 55 58

### DIFF
--- a/src/components/LinkedAccounts/ConfirmationModal/LinkedAccountToConfirm.tsx
+++ b/src/components/LinkedAccounts/ConfirmationModal/LinkedAccountToConfirm.tsx
@@ -27,11 +27,13 @@ export function LinkedAccountToConfirm({
           {account.mask}
         </P>
         <P slot='institution'>
-          {account.institution.name}
+          {account.institution?.name}
         </P>
       </VStack>
       <VStack justify='center'>
         <Checkbox
+          size='lg'
+          variant='dark'
           isSelected={isConfirmed}
           onChange={onChangeConfirmed}
           aria-label='Confirm Account Inclusion'
--- a/src/components/LinkedAccounts/ConfirmationModal/LinkedAccountToConfirm.tsx
+++ b/src/components/LinkedAccounts/ConfirmationModal/LinkedAccountToConfirm.tsx
@@ -33,7 +33,6 @@ export function LinkedAccountToConfirm({
       <VStack justify='center'>
         <Checkbox
           size='lg'
-          variant='dark'
           isSelected={isConfirmed}
           onChange={onChangeConfirmed}
           aria-label='Confirm Account Inclusion'
--- a/src/components/ui/Checkbox/Checkbox.tsx
+++ b/src/components/ui/Checkbox/Checkbox.tsx
@@ -8,17 +8,20 @@ import classNames from 'classnames'
 
 const CLASS_NAME = 'Layer__Checkbox'
 
-type CheckboxVariant = 'default' | 'dark'
+type CheckboxVariant = 'default' | 'success'
 type CheckboxSize = 'md' | 'lg'
 
 type CheckboxProps = Omit<AriaCheckboxProps, 'className'> & {
   className?: string
   variant?: CheckboxVariant
   size?: CheckboxSize
+}
+
+type CheckboxWithTooltipProps = CheckboxProps & {
   tooltip?: string
 }
 
-export function Checkbox({ children, className, variant = 'default', size = 'md', tooltip, ...props }: CheckboxProps) {
+export function Checkbox({ children, className, variant = 'default', size = 'md', ...props }: CheckboxProps) {
   const dataProperties = useMemo(() => toDataProperties({
     size,
     variant,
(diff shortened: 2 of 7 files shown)


src/components/LinkedAccounts/ConfirmationModal/LinkedAccountToConfirm.tsx
```diff
@@ -27,11 +27,13 @@ export function LinkedAccountToConfirm({
           {account.mask}
         </P>
         <P slot='institution'>
-          {account.institution.name}
+          {account.institution?.name}
         </P>
```
imkesin, line 31: If it is possible for the institution to be `undefined`, then we need to improve our typescript types and have better fallback conditions here.

src/components/ui/Checkbox/Checkbox.tsx
```diff
@@ -3,30 +3,49 @@ import { useMemo } from 'react'
...
-          <div slot='checkbox'>
-            <Check size={16} />
-          </div>
-          {node}
-        </>
-      ))}
-    </ReactAriaCheckbox>
+    <div className='Layer__checkbox-wrapper'>
+      <Tooltip disabled={!tooltip}>
+        <TooltipTrigger className='Layer__input-tooltip'>
```
imkesin, line 31: This shouldn't live within the checkbox component at all.

If you want a tooltip, on the checkbox, wrap it right when you use it.

src/components/ui/Checkbox/Checkbox.tsx
```diff
@@ -3,30 +3,49 @@ import { useMemo } from 'react'
 import { Checkbox as ReactAriaCheckbox, type CheckboxProps as AriaCheckboxProps } from 'react-aria-components'
 import { withRenderProp } from '../../utility/withRenderProp'
 import { toDataProperties } from '../../../utils/styleUtils/toDataProperties'
+import { Tooltip, TooltipTrigger, TooltipContent } from '../../Tooltip'
+import classNames from 'classnames'
 
 const CLASS_NAME = 'Layer__Checkbox'
 
-type CheckboxProps = Omit<AriaCheckboxProps, 'className'>
+type CheckboxVariant = 'default' | 'dark'
```
imkesin, line 11: This needs to be flipped.

- "Default" is the same as our buttons
- "Success" would be the green color

src/components/ui/Checkbox/checkbox.scss
```diff
@@ -2,38 +2,72 @@
...
+  }
+
+  &[data-size="lg"] [slot="checkbox"] {
+    width: 24px;
+    height: 24px;
+    border-radius: var(--border-radius-2xs);
+
+    &:hover {
+      box-shadow: 0 0 0 2px rgb(0 0 0 / 5%);
+    }
```
imkesin, line 25: This seems like it is potentially independent of the size.

src/components/ui/Checkbox/checkbox.scss
```diff
@@ -2,38 +2,72 @@
...
     display: grid;
     place-items: center;
+    width: 16px;
+    height: 16px;
+  }
+
+  &[data-size="lg"] [slot="checkbox"] {
+    width: 24px;
+    height: 24px;
+    border-radius: var(--border-radius-2xs);
```
imkesin, line 21: This could be re-written using local variables.

src/components/ui/Checkbox/checkbox.scss
```diff
@@ -2,38 +2,72 @@
...
+    box-shadow: 0 0 0 1px var(--color-info-success-bg);
+  }
+
+  &:disabled {
+    cursor: not-allowed;
+
+    &:hover {
+      border-color: var(--color-base-300);
+    }
+  }
```
imkesin, line 72: [These have their own react aria variants](https://react-spectrum.adobe.com/react-aria/Checkbox.html#styling)

src/components/ui/Checkbox/Checkbox.tsx
tomaszantas: Good idea, it's definitely better to keep these separate.
I've added `CheckboxWithTooltip` next to the `Checkbox` in the same file

src/components/ui/Checkbox/checkbox.scss
tomaszantas: You are right, thanks!

src/styles/internal_variables.scss
```diff
@@ -19,13 +17,27 @@
...
+  --checkbox-radius: var(--border-radius-3xs);
+  --checkbox-border-color: var(--color-base-300);
+  --checkbox-color-selected: var(--bg-default);
+  --checkbox-bg-selected: var(--fg-default);
+  --checkbox-border-color-selected: var(--fg-default);
+  --checkbox-bg-disabled: var(--color-base-200);
+  --checkbox-size-lg: 24px;
+  --checkbox-radius-lg: var(--border-radius-2xs);
+  --checkbox-bg-success-selected: var(--color-info-success);
+  --checkbox-border-color-success-selected: var(--color-info-success);
```
imkesin, line 41: This is partially my fault for introducing `--button-var` at this level.

This isn't what I meant by using local variables... I'll do some example in the future to better demonstrate.

src/types/linked_accounts.ts
```diff
@@ -39,7 +39,7 @@ export type LinkedAccount = {
     created_at: string
   }
   current_ledger_balance: number
-  institution: {
+  institution?: {
     name: string
     logo: string | null
   }
```
imkesin, line 45: I think `null` would be more correct here.

imkesin (approved):

src/styles/internal_variables.scss
tomaszantas: Thank you!

PR #513, opened 2025-02-27
fix: (sometimes) hide P&L Summary Chart "Indicator"

imkesin:
## Description

This change is intended to reduce confusion when a custom date range is selected.

### DIFF
--- a/src/components/ProfitAndLossChart/Indicator.tsx
+++ b/src/components/ProfitAndLossChart/Indicator.tsx
@@ -36,11 +36,9 @@ export const Indicator = ({
   const { x: animateTo = 0, width = 0 } =
     'x' in viewBox ? viewBox : emptyViewBox
   const margin = width > 12 ? 12 : 6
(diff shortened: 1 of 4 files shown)


src/components/ProfitAndLossChart/Indicator.tsx
```diff
@@ -36,11 +36,9 @@ export const Indicator = ({
   const { x: animateTo = 0, width = 0 } =
     'x' in viewBox ? viewBox : emptyViewBox
   const margin = width > 12 ? 12 : 6
-  const boxWidth = width + margin
+  const boxWidth = width + (2 * margin)
```
imkesin, line 39: Changes in this file are intended to solve a request from Justin:
- The box appeared a bit too shifted to the right
- The box was not tall enough to hit the top of the graph

AndrewAcomb (approved):
Approving with request that you nag Justin and show him on your screen

PR #516, opened 2025-03-04
fix: Modal default size is far too small

imkesin:
## Description

It seems like the Modal was unintentionally made especially small.

## Screenshots

This screenshot shows how the modal will appear (embedded within a customer's UI) at the adjusted size.

Screenshot 2025-03-03 at 4 08 02 PM


tomaszantas (approved):
I imagine that we can make the modal more flexible. Give wide limits, like `min: 20dvh, max: 90dvh` on `Layer__Dialog`, and then allow `Layer__ModalContent` to fill available space. I couldn't find straightforward adjustments in styles within few minutes, so we can merge this to fix existing issue, and then I can put another PR trying to make modal more flexible.

PR #517, opened 2025-03-05
feat!: Overhaul `LinkAccounts` component

imkesin:
## Description

This contains breaking changes - I am under the impression that this component is not widely used (or particularly useful), so I have intentionally broken the contract to give us a clean slate.


package.json
```diff
@@ -72,6 +72,7 @@
   },
   "dependencies": {
     "@floating-ui/react": "^0.26.8",
+    "@tanstack/react-form": "^1.0.0",
```
imkesin, line 75: A start of something beautiful... 🌈

src/components/LinkedAccounts/LinkedAccounts.tsx
```diff
@@ -23,9 +24,11 @@ export interface LinkedAccountsProps {
 
 export const LinkedAccounts = (props: LinkedAccountsProps) => {
   return (
-    <LinkedAccountsProvider>
-      <LinkedAccountsComponent {...props} />
-    </LinkedAccountsProvider>
+    <AccountConfirmationStoreProvider>
```
imkesin, line 27: I now am intentionally only showing the modal from the `LinkedAccounts` component.

Previously, I thought it would make sense from every screen that supports linking accounts. The `LinkAccounts` component keeps everything in the Wizard.

src/components/PlatformOnboarding/Container/linkAccountsListContainer.scss
```diff
@@ -0,0 +1,14 @@
+.Layer__LinkAccountsListContainer {
+  display: grid;
+
+  row-gap: var(--spacing-xs);
+  grid-auto-rows: 1fr;
+
+  overflow-y: auto;
+  max-block-size: max(24rem, 60dvh);
```
imkesin, line 8: This is intended to keep the component reasonably sized when dealing with many linked accounts. Most customers will not hit the threshold.

src/components/LinkedAccounts/LinkedAccounts.tsx
```diff
@@ -7,6 +7,7 @@ import { Loader } from '../Loader'
 import { Heading, HeadingSize } from '../Typography'
 import { LinkedAccountsContent } from './LinkedAccountsContent'
 import { OpeningBalanceModal } from './OpeningBalanceModal/OpeningBalanceModal'
+import { AccountConfirmationStoreProvider } from '@layerfi/components/providers/AccountConfirmationStoreProvider'
```
tomaszantas, line 10: I think this may break local development
```suggestion
import { AccountConfirmationStoreProvider } from '../../providers/AccountConfirmationStoreProvider'
```

src/components/Onboarding/Onboarding.tsx
```diff
@@ -5,6 +5,7 @@ import { LinkedAccountsProvider } from '../../providers/LinkedAccountsProvider'
 import { OnboardingStep } from '../../types/layer_context'
 import { Container } from '../Container'
 import { ConnectAccount } from './ConnectAccount'
+import { AccountConfirmationStoreProvider } from '@layerfi/components/providers/AccountConfirmationStoreProvider'
```
tomaszantas, line 8:
```suggestion
import { AccountConfirmationStoreProvider } from '../..//providers/AccountConfirmationStoreProvider'
```

tomaszantas (commented):
Looks good in overall, I spot two places where `@layerfi/*` imports were used

src/components/LinkedAccounts/LinkedAccounts.tsx
imkesin: Apologies - I have changed one of my VSCode settings to stop it from looking in the wrong directory.

src/components/PlatformOnboarding/Steps/LinkAccountsConfirmationStep.tsx
```diff
@@ -0,0 +1,139 @@
...
+}
+
+function AccountConfirmationEmptyList() {
+  const CLASS_NAME = 'Layer__AccountConfirmationEmptyList'
+
+  return (
+    <div className={CLASS_NAME}>
+      <VStack slot='center' gap='xs'>
+        <Heading size='sm' align='center'>No Accounts Need Confirmation</Heading>
+        <P variant='subtle' align='center'>You are free to skip this step!</P>
```
doneel, line 41: This strikes me as a place where we should just... close the modal / fast forward to the next screen instead of showing this.

If that's _not_ a possibility then rather not allude to some step they can skip and instead I'd rather have a title like "Accounts successfuly linked" and a subtext like "You can link more accounts at any time from the Bank Transactions Page".

src/components/PlatformOnboarding/Steps/LinkAccountsConfirmationStep.tsx
```diff
@@ -0,0 +1,139 @@
...
+  })
+
+  return (
+    <VStack gap='lg'>
+      <VStack gap='2xs'>
+        <Heading level={3} size='sm'>
+          Which accounts do you use for businesses?
+        </Heading>
+        <P variant='subtle'>
+          Please unselect any accounts you don&apos;t use for your business.
```
doneel, line 88: Can we not use quoted strings to avoid having to put in ugly html escape characters? I do not like this and don't like the risk of it leaking as raw text in the front end in the future

src/components/PlatformOnboarding/Steps/LinkAccountsLinkStep.tsx
```diff
@@ -0,0 +1,106 @@
...
+                Connect my bank
+              </Button>
+            </HStack>
+          </VStack>
+        )}
+        Container={({ children }) => (
+          <VStack>
+            <VStack gap='2xs' pbe='md'>
+              <Heading level={3} size='sm'>
+                We&apos;ve found the below accounts
```
doneel, line 52: We've found {{effectiveAccounts.length}} accounts

src/components/PlatformOnboarding/Steps/LinkAccountsLinkStep.tsx
```diff
@@ -0,0 +1,106 @@
...
+              <P variant='subtle'>
+                You&apos;ll have the chance to remove any accounts you don&apos;t use for your business in the next step.
+              </P>
+            </VStack>
+            <LinkAccountsListContainer>
+              {children}
+            </LinkAccountsListContainer>
+            <VStack pbs='xl' gap='sm'>
+              <Heading level={3} align='center'>
+                Do you use any other bank accounts or credit cards for your business?
```
doneel, line 63: > bank accounts, credit cards, or loans

We do support linking loan accounts via plaid, though coverage is bad but let's at least try to get them to link.

doneel: Scratch this, I checked and we don't have this for partner accounts. Let's not include loans here.

doneel (approved):
Looks ready! Minor suggestions only.

src/components/PlatformOnboarding/Steps/LinkAccountsConfirmationStep.tsx
imkesin: Based on the assumptions I've made at a higher level, it should not be possible to reach this state at all... but I'm writing code like this to be extra defensive.

PR #519, opened 2025-03-07
truncate dates to local dates for all pnl requests

doneel:
## Description
At the root point where we query for P&Ls, convert all dates to be local dates to standardize on _always_ making the backend handle timezones.

### DIFF
--- a/src/api/layer/profit_and_loss.ts
+++ b/src/api/layer/profit_and_loss.ts
@@ -6,20 +6,32 @@ import type {
   ProfitAndLossSummaries,
 } from '../../types/profit_and_loss'
 import { get, post } from './authenticated_http'
+import { toLocalDateString } from '../../utils/time/timeUtils'
 
-export const getProfitAndLoss = get<{
-  data?: ProfitAndLoss
-  error?: unknown
-}>(
-  ({ businessId, startDate, endDate, tagKey, tagValues, reportingBasis }) =>
-    `/v1/businesses/${businessId}/reports/profit-and-loss?start_date=${
-      startDate ? encodeURIComponent(startDate) : ''
-    }&end_date=${endDate ? encodeURIComponent(endDate) : ''}${
+type GetProfitAndLossParams = {
+  businessId: string
+  startDate: Date
+  endDate: Date
+  tagKey?: string
+  tagValues?: string
+  reportingBasis?: string
+}
+
+export const getProfitAndLoss = (apiUrl: string, accessToken: string | undefined, params: GetProfitAndLossParams) => {
+  const { businessId, startDate, endDate, tagKey, tagValues, reportingBasis } = params
+  return get<{
+    data?: ProfitAndLoss
+    error?: unknown
+  }>(({ businessId }) =>
+    `/v1/businesses/${businessId}/reports/profit-and-loss?start_date=${encodeURIComponent(
+      toLocalDateString(startDate),
+    )}&end_date=${encodeURIComponent(toLocalDateString(endDate))}${
       reportingBasis ? `&reporting_basis=${reportingBasis}` : ''
     }${tagKey ? `&tag_key=${tagKey}` : ''}${
       tagValues ? `&tag_values=${tagValues}` : ''
     }`,
-)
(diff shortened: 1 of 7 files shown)


src/hooks/useProfitAndLoss/useProfitAndLossQuery.tsx
```diff
@@ -64,8 +64,8 @@ export const useProfitAndLossQuery: UseProfitAndLossQueryReturn = (
     Layer.getProfitAndLoss(apiUrl, auth?.access_token, {
       params: {
         businessId,
-        startDate: formatISO(startDate.valueOf()),
-        endDate: formatISO(endDate.valueOf()),
+        startDate: formatISO(startDate, { representation: 'date' }),
+        endDate: formatISO(endDate, { representation: 'date' }),
```
doneel, line 68: @imkesin I've gone back on forth on whether this makes the intent sufficiently obvious that we want this to be a simple date. Do you think it's clear enough or should we create a really-explicitly-named function like `function dateToLocalDateString`

tomaszantas (approved):
LGTM! I believe that this conversion:

```tsx
formatISO(startDate, { representation: 'date' })
```

can be commonly used with various API endpoints. So we could add this function to a new file `src/utils/api.ts` and start collecting there other helpful functions related to the API.

```tsx
// src/utils/api.ts
export const toLocalDateString = (date: Date): string => formatISO(date.valueOf(), { representation: 'date' }) 
```

src/hooks/useProfitAndLoss/useProfitAndLossQuery.tsx
imkesin: This is good.

I think we should be doing the transformations of query parameters at an even deeper level (see: `/api/layer/profit_and_loss.ts`)

imkesin (approved):

src/api/layer/profit_and_loss.ts
```diff
@@ -6,20 +6,41 @@ import type {
...
+
+export const getProfitAndLoss = (apiUrl: string, accessToken: string | undefined, params: GetProfitAndLossParams) => {
+  const { businessId, startDate, endDate, month, year, tagKey, tagValues, reportingBasis } = params
+  return get<{
+    data?: ProfitAndLoss
+    error?: unknown
+  }>(({ businessId }) =>
+    `/v1/businesses/${businessId}/reports/profit-and-loss?${
+      startDate ? `start_date=${encodeURIComponent(toLocalDateString(startDate))}` : ''
+    }${endDate ? `&end_date=${encodeURIComponent(toLocalDateString(endDate))}` : ''}${
```
doneel, line 35: The API supports year/month here in addition to start_date and end_date.
We don't actually use the year/month anywhere in the components today, but I wanted to refactor this and the CSV download to 
1. Accurately reflect the API endpoint parameters
2. Share the same types

doneel: But this change isn't actually necessary / related to the timestamp clamping

src/hooks/useProfitAndLossComparison/utils.ts
```diff
@@ -99,8 +100,8 @@ function preparePeriodsBodyForDateRange(dateRange: DateRange) {
   return {
     type: 'Comparison_Date_Ranges' as const,
     date_ranges: [{
-      start_date: dateRange.startDate.toISOString(),
-      end_date: dateRange.endDate.toISOString(),
+      start_date: toLocalDateString(dateRange.startDate),
+      end_date: toLocalDateString(dateRange.endDate),
```
doneel, line 104: I would have _liked_ to do this formatting _right where the API call gets made_ but it would be messy here as
1. It's a `POST` request and we're taking in a `formattedBody` parameter, so it's weird to then do _extra_ formatting as we make the request
2. This is a polymorphic type that supports multiple request body formats, so you would only be remapping one of them, gross.

I decided to do the formatting in the `prepareFiltersBody` method, since this formatting does match the _intent_ of these methods and because it's the single place this body gets formatted now, so feels pretty safe that we have a single point of entry where we'll always do this formatting.

imkesin (approved):

PR #520, opened 2025-03-07
feat: default bookkeeping tasks to open on mobile

tomaszantas:
## Description

Open Bookkeeping tasks by default if any unresolved task found.

Important change: now we fetch all tasks since activation date to today.

## Changes

- read the whole history of tasks
- linter issues

## How this has been tested?

https://github.com/user-attachments/assets/c3754fc7-59a7-4b0c-bc63-4d07d7402282

PR #521, opened 2025-03-07
feat: add badges to date picker

tomaszantas:
## Description

Follow up to: #520 

Highlight years in Tasks DatePicker if year contains unresolved task.

## Changes

- DatePicker - highlight years and navigation arrows if `highlightYears` is set
- Tasks - highlight years if any unresolved task

## How this has been tested?

image


imkesin (approved):
(Silly Nit)

I wish there was a way to adjust the positioning of the badge. It's in the upper right relative to the inner component, but not relative to the entire button.

PR #522, opened 2025-03-07
fix: use ISO `full-date` when transforming search parameters

imkesin:
## Description

This change is inspired by #519. It will influence our cashflow statement and balance sheet queries.


doneel (approved):
Makes a lot of sense to me. This only affects query string params on `GET` requests, correct?

imkesin:
> Makes a lot of sense to me. This only affects query string params on `GET` requests, correct?

Yup, this is only used in a handful of places... but should eventually be our preferred way of working with query parameters.

PR #523, opened 2025-03-07
include negative values in table and total, but not in percentages and pie chart

doneel:
## Description
Previously we dropped negative values entirely out of the detail charts.
Now, we're going to include the negative values in the detail charts so they affect the total and show up in the table, but exclude them from percentage calculations so we don't have nonsensical numbers.
image

### DIFF
--- a/src/components/ProfitAndLossDetailedCharts/DetailedChart.tsx
+++ b/src/components/ProfitAndLossDetailedCharts/DetailedChart.tsx
@@ -41,7 +41,7 @@ export const DetailedChart = ({
     if (!filteredData) {
       return []
     }
-    return filteredData.map(x => {
+    return filteredData.map((x) => {
       if (x.hidden) {
         return {
           ...x,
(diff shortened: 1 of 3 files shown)


src/components/ProfitAndLossDetailedCharts/DetailedChart.tsx
```diff
@@ -97,272 +97,279 @@ export const DetailedChart = ({
                 <rect width='1' height='1' opacity={0.56} />
               </pattern>
             </defs>
-            {!isLoading && !noValue ? (
```
doneel, line 100: Pretty much everything below this is formatting only

src/utils/profitAndLossUtils.ts
```diff
@@ -4,13 +4,11 @@ import { ProfitAndLoss } from '../types/profit_and_loss'
...
 const doesLineItemQualifies = (item: LineItem) => {
   return !(
-    item.is_contra ||
-    item.value === undefined ||
-    item.value === null ||
-    isNaN(item.value) ||
-    item.value === -Infinity ||
-    item.value === Infinity ||
-    item.value < 0
+    item.value === undefined
```
doneel, line 7: we were filtering out everything here -> now we're pushing that responsibility to the very end when we format the chart

PR #525, opened 2025-03-08
refactor: balance sheet data loading

imkesin:
## Description

I noticed that our balance sheet was incorrectly requesting a timestamp at the start of a day.

PR #527, opened 2025-03-10
fix: broken types related to Cashflow Statement data loading

imkesin:
## Description

This was caused by me in #525.


src/api/layer/authenticated_http.ts
```diff
@@ -24,7 +24,7 @@ export const get =
     Return extends Record<string, unknown> = Record<string, unknown>,
     Params extends Record<string, ParameterValues | null | undefined> = Record<
       string,
-      ParameterValues | null | undefined
+      string | undefined
```
imkesin, line 27: This was a mistake by me on a previous PR.

A considerable number of `get` definitions do not specify the types they expect for query parameters, so I am reverting to preserve existing behavior.

src/hooks/useStatementOfCashFlow/useStatementOfCashFlow.tsx
```diff
@@ -1,72 +1,63 @@
...
+  startDate?: Date
+  endDate?: Date
+}) {
   const { data: auth } = useAuth()
-
-  const startDateString = format(
-    startOfDay(startDate),
-    'yyyy-MM-dd\'T\'HH:mm:ssXXX',
-  )
-  const endDateString = format(startOfDay(endDate), 'yyyy-MM-dd\'T\'HH:mm:ssXXX')
```
imkesin, line 34: These need to be removed - they prevent us from getting the desired logic for date range on reports.

src/hooks/useStatementOfCashFlow/useStatementOfCashFlow.tsx
```diff
@@ -1,72 +1,63 @@
...
-    if (queryKey && (isLoading || isValidating)) {
-      read(DataModel.STATEMENT_OF_CASH_FLOWS, queryKey)
-    }
-  }, [isLoading, isValidating])
-
-  useEffect(() => {
-    if (queryKey && hasBeenTouched(queryKey)) {
-      refetch()
-    }
-  }, [syncTimestamps, startDateString, endDateString])
```
tomaszantas, line 69: After removing this, the Statement of Cashflow won't be refetched automatically after ie. categorizing bank transaction. Here is the test case:

1. Visit Reports > Statement of cash flow and select ie. August 2024 - "remember" values from the Report
2. Go to Bank Transactions and categorize any transaction from August 2024
3. Go back to Reports > Statement of cash flow - without above code, SWR won't refetch data, so you will see same values as in step 1.
4. Refresh page - you should see new values

tomaszantas (approved):
Looks good to me in overall, I left one comment

PR #530, opened 2025-03-12
include structure in processed body

doneel:
## Description
Accidentally been dropping this structure parameter when we formatted the tag filter body. This was causing us to not have any structure in the request, which I just found testing locally. 
The only customer who uses this functionality hadn't bumped to a version that dropped the structure, so fortunately, no client effects.


imkesin (approved):

src/hooks/useProfitAndLossComparison/utils.ts
```diff
@@ -16,28 +17,38 @@ export function prepareFiltersBody(compareOptions: TagComparisonOption[]) {
...
+      const filter: ProfitAndLossComparisonTags = {
+        structure,
+        required_tags: [{
+          key: tagFilters.tagKey,
+          value: tagValue,
+        }],
+      }
+      return filter
+    })
+  }).filter((item): item is ProfitAndLossComparisonTags => item !== null)
```
imkesin, line 37: (For Typescript Learning)

I would express this as:

```
item is Exclude<typeof item, null>
```

PR #536, opened 2025-03-20
Fix/linter 0325

tomaszantas:
## Description

Fix eslint issues in folders:
- icons
- types
- providers
- few hooks


src/providers/BusinessProvider/BusinessProvider.tsx
```diff
@@ -259,7 +259,7 @@ export const BusinessProvider = ({
         setOnboardingStep,
         addToast,
         removeToast,
-        onError: errorHandler.onError,
+        onError: (payload: LayerError) => errorHandler.onError(payload),
```
tomaszantas, line 262: Here is the only non-autofix issue

imkesin (approved):

PR #541, opened 2025-03-25
fix: adding spinner inside button

imkesin:
## Description

I want a button to contain a spinner _and_ have zero layout shift when changing states.

### DIFF
--- a/src/components/ui/Button/button.scss
+++ b/src/components/ui/Button/button.scss
@@ -68,3 +68,20 @@
     background-color: var(--button-bg-disabled);
   }
 }
+
+.Layer__ButtonTransparentContent {
+  opacity: 0;
+  grid-row: 1 / -1;
+  grid-column: 1 / -1;
+}
+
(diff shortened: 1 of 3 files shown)


src/components/ui/Button/button.scss
```diff
@@ -68,3 +68,15 @@
...
+.Layer__ButtonTransparentContent {
+  opacity: 0;
+  grid-row: 1 / -1;
+  grid-column: 1 / -1;
+}
+
+.Layer__ButtonSpinnerContainer {
+  grid-column: 1 / -1;
+  grid-row: 1 / -1;
+  place-self: center;
```
tomaszantas, line 81: I think the spinner is not centered vertically. Probably adding padding or static height will improve that:
```
  padding-top: 3px;
  // Or:
  height: 18px;
```

image

image

tomaszantas (approved):
LGTM! the spinner position could be slighlty improved

PR #555, opened 2025-04-09
fix: prevent jumping layout in bookkeeping overview

tomaszantas:
## Description

This PR adds hook observing `upper` element size and adjusts the page scroll to keep the `target` in the same spot of the viewport.

The usage:

```tsx
const { upperContentRef, targetElementRef, upperElementInFocus } = useKeepInMobileViewport()

return (
  <>
   <div ref={upperContentRef} onPointerEnter={() => upperElementInFocus.current = true}>
    {...}
   </div>

    <div ref={targetElementRef} onPointerEnter={() => upperElementInFocus.current = false}>
       {...} 
    </div>
  </>
)
```

---

To discuss:

We can leverage this into more reusable component, something like this:

```tsx
import { StickyMobileScroll } from '../components/StickyMobileScroll'

return (
   <StickyMobileScroll upperElement={<div>...</div>}>
      <div>Target element</div>
  </StickyMobileScroll>
}
```

...and add prop for which we disable behavior.

However, making this as a local hook dedicated just for Bookkeeping Overview also makes sense to me at this point, because I don't foresee other use cases.

## How this has been tested?

https://github.com/user-attachments/assets/473b3cac-46af-4b83-9d9e-7b09e3181402

### DIFF
--- a/src/hooks/useKeepInMobileViewport.tsx
+++ b/src/hooks/useKeepInMobileViewport.tsx
@@ -1,0 +1,74 @@
+import { useEffect, useRef } from 'react'
+
+export const useKeepInMobileViewport = () => {
+  const upperContentRef = useRef<HTMLDivElement>(null)
+  const targetElementRef = useRef<HTMLDivElement>(null)
+  const lastKnownSizeRef = useRef<number | null>(null)
+  const lastKnownTargetPositionRef = useRef<number | null>(null)
+  const upperElementInFocus = useRef<boolean | null>(false)
+
+  useEffect(() => {
+    if (!upperContentRef.current || !targetElementRef.current) return
+
+    lastKnownSizeRef.current = upperContentRef.current.getBoundingClientRect().height
+    lastKnownTargetPositionRef.current = targetElementRef.current.getBoundingClientRect().top
+
+    // Use observer to detect upper element content changes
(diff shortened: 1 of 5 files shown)


src/hooks/useKeepInMobileViewport.tsx
```diff
@@ -0,0 +1,74 @@
+import { useEffect, useRef } from 'react'
+
+export const useKeepInMobileViewport = () => {
+  const upperContentRef = useRef<HTMLDivElement>(null)
+  const targetElementRef = useRef<HTMLDivElement>(null)
+  const lastKnownSizeRef = useRef<number | null>(null)
+  const lastKnownTargetPositionRef = useRef<number | null>(null)
+  const upperElementInFocus = useRef<boolean | null>(false)
```
imkesin, line 8: I don't think this one can be `null` ever based on the current implementation

imkesin (approved):
I find `ResizeObserver`-based solutions to be very difficult to reason about (skill issue?).

I cannot get these changes to work properly when starting from a non-mobile layout and the resizing down to mobile.

tomaszantas:
Using `touchEvent` didn't work well, so I changed to `onClick`. Now it should be much more reliable. Also, I moved the hook next to the Bookkeeping overview.

PR #557, opened 2025-04-10
fix: hide years without any active bookkeeping period

imkesin:
## Description

If a business did not purchase bookkeeping in 2023, then we should not see any "2023" tab for them.


AndrewAcomb (approved):
nice

PR #560, opened 2025-04-11
fix: invalidate periods after mutating a task

imkesin:
## Description

When testing the new Bookkeeping UI, I was reminded by @AndrewAcomb that revalidation (or invalidation) may not be working properly.

This is a massive overhaul to get this into a better spot.


src/components/Tasks/TasksListItem.tsx
```diff
@@ -188,9 +201,11 @@ export const TasksListItem = ({
...
                     variant={ButtonVariant.secondary}
                     onClick={() => {
-                      submitResponseToTask(task.id, userResponse)
-                      setIsOpen(false)
-                      goToNextPageIfAllComplete(task)
+                      void handleSubmitUserResponseForTask({ taskId: task.id, userResponse })
+                        .then(() => {
+                          setIsOpen(false)
+                          goToNextPageIfAllComplete(task)
+                        })
```
doneel, line 208: 👍

src/hooks/bookkeeping/periods/tasks/useSubmitResponseForTask.ts
```diff
@@ -0,0 +1,95 @@
...
+  const { trigger: originalTrigger } = mutationResponse
+
+  const stableProxiedTrigger = useCallback(
+    async (...triggerParameters: Parameters<typeof originalTrigger>) => {
+      const result = await originalTrigger(...triggerParameters)
+
+      if (result) {
+        void mutate(key => withSWRKeyTags(
+          key,
+          tags => tags.includes(BOOKKEEPING_PERIODS_TAG_KEY),
```
doneel, line 76: Is this `tags` pattern the answer for managing what needs downstream invalidation?

src/hooks/bookkeeping/periods/useBookkeepingPeriods.ts
```diff
@@ -95,11 +99,24 @@ export function useBookkeepingPeriods() {
...
+  return new Proxy(swrResponse, {
+    get(target, prop) {
+      if (prop === 'isLoading') {
+        return isLoadingBookkeepingStatus || swrResponse.isLoading
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+      return Reflect.get(target, prop)
+    },
+  })
```
doneel, line 121: This looks like scary witchcraft to me and I want to understand this pattern (as opposed to like... explicitly returning an object with the `swrRsponse` and a loading state.

doneel (approved):

src/hooks/bookkeeping/periods/tasks/useSubmitResponseForTask.ts
imkesin: Yup!

This forces any SWR with `{ tags: [  '#bookkeeping-periods'  ] }` to revalidate.

src/hooks/bookkeeping/periods/useBookkeepingPeriods.ts
imkesin: It is witchcraft.

For an SWR response, all of the members of the object - `data`, `isLoading`, etc. - are actually getters with a side effect that marks them as "tracked".

If you can avoid spreading or destructuring the object, then you reduce the reactivity. This is especially useful for avoiding silly re-renders on `isValidating`.

In this particular case, the query is dependent on another query - it doesn't make a network request until it knows that we're in a proper bookkeeping status. From the perspective of the caller, this should also be counted as "loading" time.

PR #561, opened 2025-04-12
Daniel/onboarding copy tweaks

doneel:
### Before connecting first account
image

### After connecting first account
image

Image

Image

Image


src/styles/hover-menu.scss
```diff
@@ -70,6 +70,7 @@
           font-family: inherit;
           font-feature-settings: inherit;
           font-weight: 500;
+          width: 100%;
```
doneel, line 73: This is unrelated, but fixes the issue where clicking on the 'unlink account' button from the three-dot dropdown seems to not work because you're not clicking the text.

jmeretab (approved):

PR #565, opened 2025-04-15
Daniel/lightweight form by default

doneel:
Simplified form
image

Optional back button in the link accounts component (only used when embedded within onboarding form)
image

### DIFF
--- a/src/components/BusinessForm/BusinessForm.tsx
+++ b/src/components/BusinessForm/BusinessForm.tsx
@@ -13,13 +13,16 @@ export type BusinessFormStringOverrides = {
   saveButton?: string
 }
 
+export type BusinessFormVariant = 'lightweight' | 'full'
+
 export type BusinessFormProps = {
   stringOverrides?: BusinessFormStringOverrides
   onSuccess?: () => void
+  variant?: BusinessFormVariant
 }
 
-export const BusinessForm = ({ stringOverrides, onSuccess }: BusinessFormProps) => {
-  const { form, submitError, isFormValid } = useBusinessForm({ onSuccess })
+export const BusinessForm = ({ stringOverrides, onSuccess, variant = 'lightweight' }: BusinessFormProps) => {
+  const { form, submitError, isFormValid } = useBusinessForm({ onSuccess, variant })
 
   const { isSubmitting } = form.state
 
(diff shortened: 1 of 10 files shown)


src/components/BusinessForm/BusinessForm.tsx
```diff
@@ -13,13 +13,16 @@ export type BusinessFormStringOverrides = {
   saveButton?: string
 }
 
+export type BusinessFormVariant = 'lightweight' | 'full'
+
 export type BusinessFormProps = {
   stringOverrides?: BusinessFormStringOverrides
   onSuccess?: () => void
+  variant?: BusinessFormVariant
```
imkesin, line 21: What's the rationale for maintaining both paths?

src/components/PlatformOnboarding/Steps/BusinessInfoStep.tsx
```diff
@@ -1,18 +1,24 @@
...
-export const BusinessInfoStep = ({ title = defaultTitle, onNext }: BusinessInfoStepProps) => {
+export const BusinessInfoStep = ({ title = defaultTitle, onNext, onBack }: BusinessInfoStepProps) => {
   return (
     <>
       <Heading>{title}</Heading>
-      <BusinessForm onSuccess={onNext} />
+      <BusinessForm onSuccess={onNext} stringOverrides={{ saveButton: '' }} />
+      <div className='Layer__platform-onboarding__button-row'>
+        <Button onClick={onBack} variant={ButtonVariant.secondary}>Back</Button>
+        <Button onClick={() => document.querySelector('.Layer__business-form')?.dispatchEvent(new Event('submit', { bubbles: true }))}>Next</Button>
```
imkesin, line 20: This seems crazy - invisible button + a raw lookup to trigger an event?

src/components/BusinessForm/BusinessForm.tsx
doneel: Most platforms will have a bunch of this information already so most of the time, we shouldn't need to ask, but there are probably platforms that _don't_ so I want an option for those situations.

src/components/PlatformOnboarding/Steps/BusinessInfoStep.tsx
doneel: Yes you're right sorry don't merge this I will fix.

PR #566, opened 2025-04-16
upgrade: `recharts` on latest version

imkesin:
## Description

This is motivated by fixing a customer issue with `recharts` loading. It is still unclear if the version upgrade will have any notable effect.


tomaszantas (approved):
LGTM!

PR #577, opened 2025-04-24
feat: categorize dropdown

tomaszantas:
## Description

Categorize dropdown.


src/components/Tooltip/Tooltip.tsx
```diff
@@ -86,7 +88,7 @@ export const TooltipContent = forwardRef<
     <FloatingPortal>
       <div
         ref={ref}
-        className={className}
+        className={classNames('Layer__tooltip', className)}
```
tomaszantas, line 91: Fyi, `Layer__tooltip` class is added manually to `<Tooltip .../>` component in many places. I'm going refactor this with another PR once this is merged.

src/components/CategorySelect/CategorySelect.tsx
```diff
@@ -257,65 +23,34 @@ export const CategorySelect = ({
...
-          } satisfies GroupBase<CategoryOption>,
-      ]
-      : []
+  const allKeys = useMemo(() => {
+    const categoriesKeysMap = getKeysMap(allCategories.flatMap(group => group.options))
+    const matchKeysMap = getKeysMap(matches ?? [])
 
-  const categoryOptions = flattenCategories(categories)
+    return new Map([...categoriesKeysMap, ...matchKeysMap])
+  }, [allCategories, matches])
```
tomaszantas, line 38: ComboBox's `onSelectionChange` returns just a `Key` (for example `id`), but we need whole object in few places, ie. to properly format matches and splits. This trick stores `CategoryOption`s in map, component provides whole object in `onChange` callback.

src/components/CategorySelect/components/SuggestionsList.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <ListSection aria-label='Suggestions'>
+      <Header slot='header'>
+        <Text size={TextSize.xs} status='disabled'>Suggestions</Text>
+      </Header>
+      {suggestions.map((option, index) => (
+        <ListBoxItem
+          key={index}
+          id={`suggestion-${option.payload.id}`}
```
tomaszantas, line 31: The `suggestion-` prefix resolves strange ComboBox behavior. Without prefix, all categories are rendered in `Suggestions` section:

image


The expected result:

image

Probably this happens due to duplicated id - another item in "All categories" has same category id

src/components/CategorySelect/components/ListItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+            </TooltipContent>
+          </Tooltip>
+        )}
+      </div>
+
+      <span slot='account'>
+        <Text size={TextSize.sm} status='disabled' ellipsis>{accountName}</Text>
+      </span>
+
+      {isSelected(option, selected) && (
```
tomaszantas, line 50: `<ListBoxItem />` provides `isSelected` prop:

```
<ListBoxItem>
 {({ isSelected }) => (
    ...
 )}
</ListBoxItem>
```

However, it doesn't allow to select two items at the same time - one from suggestions list and another from regular list.

src/components/CategorySelect/CategorySelect.tsx
```diff
@@ -327,50 +62,42 @@ export const CategorySelect = ({
...
+          <ListBox>
+            <MatchesList matches={matches} selected={value} />
+
+            <SuggestionsList suggestions={suggestions} categories={categories} selected={value} />
+
+            <ListSection>
+              <ListBoxItem isDisabled={true}>
+                <Header slot='header'>
+                  <Text size={TextSize.xs}>All categories</Text>
+                </Header>
```
imkesin, line 84: I believe this will cause an accessibility warning:

```
A `textValue` prop is required for <ListBoxItem> elements with non-plain text children in order to support accessibility features such as type to select.
```

src/components/CategorySelect/components/CategoriesList.tsx
```diff
@@ -0,0 +1,79 @@
...
+            level={level + 1}
+            indentationLevel={level}
+            showTooltips={showTooltips}
+            selected={selected}
+          />
+        ))}
+
+        <ListBoxItem isDisabled={true} textValue='' className='Layer__category-select__separator'>
+          <span slot='separator-line' />
+        </ListBoxItem>
```
imkesin, line 41: The "hack" of disabling a bunch of items to trap click events seems like it has gone pretty far... it's spreading a very specific concern beyond a certain scope.

I was thinking of doing something like this at the top-level

```
const refCallback = useCallback((ref: HTMLDivElement | null) => {
    if (ref) {
      ref.addEventListener('click', stopPropagation)
    }

    return () => {
      if (ref) {
        ref.removeEventListener('click', stopPropagation)
      }
    }
  }, [stopPropagation])
```

... and then including a comment to specifically describe why.

src/components/CategorySelect/components/ListItem.tsx
tomaszantas: "Suggestions" may contain same items as "All categories" (at least in terms of account ID). If I apply same `id={option.payload.id}` prop in both "All categories" and "Suggestions", then "Suggestions" list for some reasons breaks and contains items from "All categories" after that item:

Broken suggestions list:
image

...here what we have further on the list:
image

So using duplicated IDs breaks Combobox. So I added prefixes to matches and suggestions:

```tsx
<ListBoxItem  id={`suggestion-${option.payload.id}`} (...) />

<ListBoxItem  id={`match-${option.payload.id}`} (...) />
```

But then, `isSelected` from `<ListBoxItem />` won't select multiple items on the list: one on suggestions list and another on all categories list. That's why I added custom `isSelected` function

src/components/BankTransactionRow/BankTransactionRow.tsx
```diff
@@ -59,7 +54,7 @@ export const extractDescriptionForSplit = (category: CategoryWithEntries | null)
     return ''
   }
 
-  return category.entries.map(c => c.category.display_name).join(', ')
+  return category?.entries.map(c => c.category.display_name).join(', ')
```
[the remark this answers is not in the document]
tomaszantas: Good point! Thanks!

src/components/CategorySelect/components/CategoriesList.tsx
tomaszantas: I played with `stopPropagation` before, and it didn't help, but I could miss something. I explored also `ListBoxItem`'s `onAction`, but `isDisabled={true}` works better and it is simpler.

PR #578, opened 2025-04-24
[LAY-1298] fix: `LinkedAccountPill` should not overflow container

imkesin:
## Description

**Linear**: [LAY-1298](https://linear.app/layerfi/issue/LAY-1298/repair-connection-button-doesnt-overflow-on-mobile-devices)

This fix is very involved because the "correct" way to do popover behavior is through a portal - not with absolute positioning. In order to have scroll in the x-direction, we are constraining the overflow in the y-direction (gets rid of the second scroll bar). This also prevents elements that originating from within the box from "bleeding" outside.

The behavior and styles are not perfectly identical.

### Screenshot (Corrected)

Screenshot 2025-04-24 at 3 14 10 PM


stylelint.config.mjs
imkesin, line 1: I finally turned-on `stylelint` locally.

doneel (approved):
It _looks_ good - I would love a 5 minute Ted Talk on the how "portal" concept replaces absolute positioning.

PR #582, opened 2025-04-25
[LAY-1294] fix: downloading bank transactions should respect filters

imkesin:
## Description

**Linear**: [LAY-1294](https://linear.app/layerfi/issue/LAY-1294/txn-export-not-respecting-search-filters-applied)

The download was not receiving any of the selected filters properly. This is still not a complete fix because the server does not respect all query parameters for a download.


doneel (approved):
LGTM

PR #600, opened 2025-05-03
[LAY-1330] fix: button overflow from confirmation modal

imkesin:
## Description

The button in the modal was defaulting to a size of medium - not icon-specific sizing - so it was overflowing its boundaries.

### DIFF
--- a/stylelint.config.mjs
+++ b/stylelint.config.mjs
@@ -36,6 +36,7 @@ export default {
     'custom-property-pattern': '^[a-z0-9-]+$',
     'no-duplicate-selectors': true,
     'no-descending-specificity': null,
+    'declaration-block-no-redundant-longhand-properties': null,
 
     '@stylistic/declaration-colon-space-after': 'always-single-line',
     '@stylistic/indentation': 2,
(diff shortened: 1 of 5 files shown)


stylelint.config.mjs
```diff
@@ -36,6 +36,7 @@ export default {
     'custom-property-pattern': '^[a-z0-9-]+$',
     'no-duplicate-selectors': true,
     'no-descending-specificity': null,
+    'declaration-block-no-redundant-longhand-properties': null,
```
sarahraines, line 39: For my own edification - which property is now violating this rule?

sarahraines (approved):
LGTM!

stylelint.config.mjs
imkesin: This is purely based on my own CSS preferences.

The most common situation is that I want to do:
```
padding-block: var(--size-md);
padding-inline: var(--size-sm);
```

There's a joke that CSS is a "write-only" language, since it's almost always easier to restart than understand some else's incantations. I consider this to be a bit easier to scan.

The rule change doesn't prevent us from shorthand syntax (if you like that).

PR #602, opened 2025-05-05
feat: Create new `useCreateChildAccount` hook to create child ledger account from parent

sarahraines:
## Description
This PR serves as the frontend implementation for a new API method that allows you to create a new child account (i.e., category) under a parent ledger account specified in the URL. 

### URL
`/v1/businesses/${businessId}/ledger/accounts/${accountId}/create-child-account`

### URL Parameters

| Parameter     | Type     | Required | Description                          |
|---------------|----------|----------|--------------------------------------|
| `businessId`  | `string` | ✅       | The ID of the business               |
| `accountId`   | `string` | ✅       | The ID of the parent ledger account  |

### Request Body

```
{
  name: string;
  stable_name?: {
    type: 'StableName';
    stable_name: string;
  };
}
```

PR #605, opened 2025-05-06
refactor!: decouple the default `categories` from the LayerContext

imkesin:
## Description

This is a fast-follow to #602 which will invalidate any query tagged with `#categories`.

This is (technically) a breaking change because `useLayerContext` is exposed at the top-level. We are moving away from exposing hook and contexts at the top-level already.

### DIFF
--- a/src/hooks/categories/useCategories.ts
+++ b/src/hooks/categories/useCategories.ts
@@ -1,7 +1,10 @@
 import useSWR from 'swr'
 import { useLayerContext } from '../../contexts/LayerContext'
 import { useAuth } from '../useAuth'
-import { getCategories } from '../../api/layer/categories'
+import {
+  getCategories,
+  type CategoriesListMode,
+} from '../../api/layer/categories'
 
 export const CATEGORIES_TAG_KEY = '#categories'
 
@@ -14,7 +17,7 @@ function buildKey({
   access_token?: string
   apiUrl?: string
   businessId: string
-  mode?: 'ALL'
+  mode?: CategoriesListMode
 }) {
   if (accessToken && apiUrl) {
     return {
@@ -28,7 +31,7 @@ function buildKey({
 }
 
 type UseCategoriesOptions = {
-  mode?: 'ALL'
+  mode?: CategoriesListMode
 }
 
 export function useCategories({ mode }: UseCategoriesOptions = {}) {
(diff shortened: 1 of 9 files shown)


src/hooks/categories/useCategories.ts
```diff
@@ -4,34 +4,42 @@ import { useAuth } from '../useAuth'
...
 function buildKey({
   access_token: accessToken,
   apiUrl,
   businessId,
+  mode,
 }: {
   access_token?: string
   apiUrl?: string
   businessId: string
+  mode?: 'ALL'
```
sarahraines, line 17: NIT: Can we pull out the different category modes into an enum, and also offer an alternative to `ALL` so that the enum documents the different cases (even if we don't actually pass that alternative value to the API)?

src/hooks/categories/useCategories.ts
```diff
@@ -4,34 +4,42 @@ import { useAuth } from '../useAuth'
 import { getCategories } from '../../api/layer/categories'
 
 export const CATEGORIES_TAG_KEY = '#categories'
```
sarahraines, line 6: For the future -

I wonder if we might want to use symbols for tag keys so that we can enforce uniqueness even if the same value is chosen, and also ensure that all call sites get updated if we want to change the value.

sarahraines (approved):
Looks great! Suggested one small improvement and one potential change for the future.

src/hooks/categories/useCategories.ts
```diff
@@ -4,34 +4,42 @@ import { useAuth } from '../useAuth'
...
-      mode: 'ALL',
+      mode,
       tags: [CATEGORIES_TAG_KEY],
     } as const
   }
 }
 
-export function useAllCategories() {
+type UseCategoriesOptions = {
+  mode?: 'ALL'
```
sarahraines, line 31: NIT: Can we pull out the different category modes into an enum, and also offer an alternative to `ALL` so that the enum documents the different cases (even if we don't actually pass that alternative value to the API)?

src/hooks/categories/useCategories.ts
imkesin: Symbols are being serialized by `swr` at some deep level (before comparing), so we won't get uniqueness by using them.

src/hooks/categories/useCategories.ts
imkesin: I've included them in a revision commit.

src/hooks/categories/useCategories.ts
imkesin: ([Reference](https://github.com/vercel/swr/pull/1429#issue-988378882))

sarahraines: Sad - thanks for the reference :(

PR #615, opened 2025-05-13
fix: revalidate {Bank|External}Accounts when categorizing or matching

imkesin:
## Description

When we categorize or match a transaction, the ledger account balance at the top of the page should be updated to reflect the change.

### DIFF
--- a/src/hooks/bookkeeping/periods/tasks/useSubmitResponseForTask.ts
+++ b/src/hooks/bookkeeping/periods/tasks/useSubmitResponseForTask.ts
@@ -67,19 +67,18 @@ export function useSubmitUserResponseForTask() {
   const { trigger: originalTrigger } = mutationResponse
 
   const stableProxiedTrigger = useCallback(
-    async (...triggerParameters: Parameters<typeof originalTrigger>) => {
-      const result = await originalTrigger(...triggerParameters)
-
-      if (result) {
-        void mutate(key => withSWRKeyTags(
-          key,
-          tags => tags.includes(BOOKKEEPING_PERIODS_TAG_KEY),
-        ))
-      }
-
-      return result
-    },
-    [originalTrigger, mutate],
+    async (...triggerParameters: Parameters<typeof originalTrigger>) =>
+      originalTrigger(...triggerParameters)
+        .finally(() => {
+          void mutate(key => withSWRKeyTags(
+            key,
+            tags => tags.includes(BOOKKEEPING_PERIODS_TAG_KEY),
+          ))
+        }),
+    [
+      originalTrigger,
+      mutate,
+    ],
   )
 
   return new Proxy(mutationResponse, {
--- a/src/hooks/bookkeeping/periods/tasks/useSubmitResponseForTask.ts
+++ b/src/hooks/bookkeeping/periods/tasks/useSubmitResponseForTask.ts
@@ -67,14 +67,16 @@ export function useSubmitUserResponseForTask() {
   const { trigger: originalTrigger } = mutationResponse
 
   const stableProxiedTrigger = useCallback(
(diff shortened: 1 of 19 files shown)


src/hooks/useBankTransactions/useAugmentedBankTransactions.tsx
```diff
@@ -236,144 +231,113 @@ export const useAugmentedBankTransactions = (
...
+    })
+
+    void mutate(updatedData, { revalidate: false })
+  }
+
+  const { trigger: categorizeBankTransaction } = useCategorizeBankTransaction({
+    mutateBankTransactions: mutate,
+  })
+
+  const categorizeWithOptimisticUpdate = async (
```
imkesin, line 251: I have a hunch that we are doing something wrong with optimistic updates... but I need more time to make a bigger change with how these are handled.

sarahraines (approved):
Amazing! 🍾

src/hooks/useBankTransactions/useAugmentedBankTransactions.tsx
```diff
@@ -236,144 +231,113 @@ export const useAugmentedBankTransactions = (
...
-    return Layer.categorizeBankTransaction(apiUrl, auth?.access_token, {
-      params: { businessId, bankTransactionId: id },
-      body: newCategory,
+    return categorizeBankTransaction({
+      bankTransactionId,
+      ...newCategory,
     })
-      .then(({ data: newBT, errors }) => {
-        if (newBT) {
-          newBT.recently_categorized = true
```
tomaszantas, line 257: Dropping `recently_categorized` removes collapsing animation:

https://github.com/user-attachments/assets/b0cbf2ae-1a37-4a2b-8f2b-176601987054

https://github.com/user-attachments/assets/fdaa3b66-84d1-49e9-b326-c9cdf13c9b08

Current behaviour:
https://github.com/user-attachments/assets/cca2dcb3-cb4e-41ca-a69c-0da6efd8120e

---

It could be fixed by adding `recently_categorized` to `updateOneLocal` (in both categorize and match actions):

```tsx
if (targetedTransaction) {
     updateOneLocal({
      ...targetedTransaction,
       error: error instanceof Error ? error.message : 'An unknown error occurred',
       processing: false,
       recently_categorized: true, // <-- ADDED
    })
}
```

...but it may work only in "Categorized" tab view, because new logic triggers fetching bank transactions, and API response may come before animation completes.
This one probably could be fixed with moving callback call to `removeAfterCategorize()`.

But if you want to remove `recently_categorized` from `useAugmentedBankTransactions.tsx`, then you can remove also `removeAfterCategorize()`, and few `useEffect()`s from `BankTransactionListItem`, `BankTransactionMobileListItem`.

src/hooks/bookkeeping/periods/tasks/useSubmitResponseForTask.ts
```diff
@@ -67,19 +67,18 @@ export function useSubmitUserResponseForTask() {
...
-    },
-    [originalTrigger, mutate],
+    async (...triggerParameters: Parameters<typeof originalTrigger>) =>
+      originalTrigger(...triggerParameters)
+        .finally(() => {
+          void mutate(key => withSWRKeyTags(
+            key,
+            tags => tags.includes(BOOKKEEPING_PERIODS_TAG_KEY),
+          ))
+        }),
```
sarahraines, line 77: Checking one thing! The commit message says `fix: mutations should not block on cache invalidations`, but do you mean the opposite, that we should do cache invalidation for related keys even if the mutation fails?

imkesin: (This was resolved offline)

PR #616, opened 2025-05-13
fix: text within a primary button needs a more specific selector

imkesin:
## Description

This is a symptom of selector specificity being too low across the app. If we have any text, we need to be setting the `color` property as close as possible to the text.

### Screenshot

Screenshot 2025-05-13 at 3 11 29 PM

### DIFF
--- a/src/styles/button.scss
+++ b/src/styles/button.scss
@@ -5,14 +5,16 @@
   font-family: var(--font-family);
   font-weight: var(--font-weight-normal);
   font-variant-numeric: lining-nums proportional-nums;
-  font-feature-settings: 'cv10' on,
-    'cv05' on,
-    'cv08' on,
-    'ss03' on;
+  font-feature-settings:
+    "cv10" on,
+    "cv05" on,
+    "cv08" on,
+    "ss03" on;
   cursor: pointer;
   min-height: 36px;
   box-sizing: border-box;
-  box-shadow: 0 0 0 1px var(--color-base-300),
+  box-shadow:
+    0 0 0 1px var(--color-base-300),
     0 1px 1px 0 rgb(0 0 0 / 4%),
     0 2px 3px 0 rgb(0 0 0 / 4%),
     0 3px 4px 0 rgb(0 0 0 / 2%),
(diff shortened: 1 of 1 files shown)


src/styles/button.scss
imkesin, line 1: Most of the changes in this file are related to `stylelint`

src/styles/button.scss
```diff
@@ -75,6 +77,10 @@
     .Layer__btn-text {
       padding: 0 var(--spacing-md);
     }
+
+    .Layer__btn-content {
+      column-gap: 0;
+    }
```
imkesin, line 83: This was moved-up due to a duplicate selector.

src/styles/button.scss
```diff
@@ -196,6 +192,12 @@
       }
     }
   }
+
+  > .Layer__btn-content {
+    > .Layer__btn-text {
+      color: var(--btn-color-primary);
+    }
+  }
```
imkesin, line 200: This is the key change.

Our current state of buttons demands some major "rip-n-replace".

imkesin: Changes like these are obviously not sustainable.

src/styles/button.scss
```diff
@@ -380,7 +373,8 @@
 
     &:focus,
     :active {
-      box-shadow: 0 0 0 1px var(--color-base-0),
+      box-shadow:
```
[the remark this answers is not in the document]
imkesin: Good bot - I like the fix.

src/styles/button.scss
sarahraines: I see, do we need to get as granular as to specify a direct child relationship here rather than a descendant relationship?

sarahraines (approved):

src/styles/button.scss
imkesin: (In my opinion) Descendent relationships are an anti-pattern.

If you need nested selectors, then they better be direct child selectors.

sarahraines: This make senses, but I wonder if we ever restructure the button that we'll break these relationships. 

For example, in this case I might want to specify that any text inside the button uses the primary button text color (regardless of whether `.Layer__btn-content` exists), and so would usually reach for something like:

```
.Layer__btn {
    .Layer__btn-text {
      color: var(--btn-color-primary);
    }
}
```

sarahraines: I don't have strong opinions on this one.

PR #617, opened 2025-05-13
fix: `Filters` label is partially cutoff

imkesin:
## Description

I believe that the root cause of the issue is the use of `width: 100%`, but it is easier to force a grid layout and move on.

## Screenshot

Screenshot 2025-05-13 at 4 07 36 PM

### DIFF
--- a/src/styles/profit_and_loss_detailed_charts.scss
+++ b/src/styles/profit_and_loss_detailed_charts.scss
@@ -5,7 +5,7 @@
   align-items: flex-start;
   padding: var(--spacing-xl);
   padding-bottom: var(--spacing-lg);
-  box-shadow: 0px -1px 0px 0px var(--color-base-300) inset;
+  box-shadow: 0 -1px 0 0 var(--color-base-300) inset;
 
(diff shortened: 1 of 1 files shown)


src/styles/profit_and_loss_detailed_charts.scss
```diff
@@ -314,6 +312,9 @@ header.Layer__profit-and-loss-detailed-charts__header--tablet {
       justify-content: space-between;
 
       .chart-field {
+        display: flex;
+        flex-direction: column;
+
```
imkesin, line 317: This was moved into here because the selector was duplicated.

src/styles/profit_and_loss_detailed_charts.scss
```diff
@@ -99,7 +100,9 @@ header.Layer__profit-and-loss-detailed-charts__header--tablet {
   }
 
   .filters {
-    display: flex;
+    display: grid;
+    grid-template-columns: auto minmax(0, 1fr);
+
```
imkesin, line 105: This is the fix

sarahraines (approved):

PR #621, opened 2025-05-15
fix: categorization must invalidate bank account tags

imkesin:
## Description

I made a [bad change here](https://github.com/Layer-Fi/layer-react/pull/615/files#diff-8e8abec8c4e071254eb9be7d2bd419ed2463aa86cd0cb3e7d64af9f84cddf0eaR82) (autocomplete-related) on a previous PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@layerfi/components",
-  "version": "0.1.99-alpha.3",
+  "version": "0.1.99",
   "description": "Layer React Components",
   "main": "dist/cjs/index.cjs",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@layerfi/components",
-  "version": "0.1.99-alpha.3",
```
sarahraines, line 3: Just checking - did you mean to promote components to 0.1.99 in this PR?

imkesin: Yeah, I'm calling this the stable release and folding everything in.

PR #624, opened 2025-05-17
fix: remove `z-index` on our view headers

imkesin:
## Description

By arbitrarily increasing the z-index of our headings, we can interfere with items that are intended to overlap the header.

### Example

Screenshot 2025-05-16 at 6 25 22 PM

### DIFF
--- a/src/styles/view.scss
+++ b/src/styles/view.scss
@@ -2,12 +2,12 @@
   position: relative;
 
   .Layer__view-main {
-    padding: var(--spacing-lg);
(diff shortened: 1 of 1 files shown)


src/styles/view.scss
imkesin, line 1: Almost all of the changes in the file are re-ordering based on a new stylelint rule.

imkesin (commented):
I would like to try to understand why we messed with the `z-index` in the first place.

doneel (approved):

PR #645, opened 2025-06-05
fix: try to kill a couple `useEffect` calls

imkesin:
## Description

I found a few uses of `useEffect` that were easier to kill.


sarahraines (approved):
LGTM!

PR #646, opened 2025-06-06
feat: add data-loading for customers and vendors

imkesin:
## Description

This PR should be entirely inert.

### Side Note

Vendor selection within `Bills` isn't battle tested - it doesn't account for 100s of vendors (which will require server-side filtering).


src/features/vendors/api/useListVendors.ts
```diff
@@ -0,0 +1,126 @@
...
+      accessToken,
+      apiUrl,
+      businessId,
+      cursor: previousPageData?.meta.pagination.cursor ?? undefined,
+      tags: [VENDORS_TAG_KEY],
+    } as const
+  }
+}
+
+class ListVendorsSWRResponse {
```
sarahraines, line 66: I wonder if we can create a generic class for this that takes the API response type as a parameter `T`, where in this case T is `ListVendorsRawResult`

PR #649, opened 2025-06-09
feat: confirmation state and table summary stats for CSV upload

sarahraines:
## Description
This PR adds to the UX of CSV upload, including:
1. Having table summary stats for the uploaded CSV (e.g., Total transactions: 3, Invalid transactions: 1)
2. Having a success state

Along the way, augmented the `Wizard` component to give more flexibility around navigating between steps and syncing the current step with the parent using `onStepChange`. 

Screenshot 2025-06-09 at 2 03 19 PM

### DIFF
--- a/src/components/Wizard/Wizard.tsx
+++ b/src/components/Wizard/Wizard.tsx
@@ -5,7 +5,6 @@ import {
   createContext,
   useCallback,
   useContext,
-  useEffect,
   useMemo,
   useState,
 } from 'react'
@@ -25,7 +24,8 @@ function useWizardStep({
 
   const goToStep = useCallback((stepIndex: number) => {
     setActiveStepIndex(stepIndex)
-  }, [])
+    onStepChange?.(stepIndex)
+  }, [onStepChange])
 
   const next = useCallback(async () => {
     if (activeStepIndex === stepCount - 1) {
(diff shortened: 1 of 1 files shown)


src/components/Wizard/Wizard.tsx
```diff
@@ -5,41 +5,64 @@ import {
...
   onComplete?: () => Awaitable<void>
 }) {
   const stepCount = steps.length
   const [activeStepIndex, setActiveStepIndex] = useState(0)
 
-  const next = useCallback(async () => {
-    setActiveStepIndex(stepIndex => Math.min(stepIndex + 1, stepCount - 1))
+  const goToStep = useCallback((stepIndex: number) => {
+    setActiveStepIndex(stepIndex)
+  }, [])
```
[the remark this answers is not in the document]
sarahraines: Don't need this, the `stepIndex` should be a number that comes from an enum that maps to the steps shown.

src/components/Wizard/Wizard.tsx
```diff
@@ -5,41 +5,64 @@ import {
...
-  const previous = useCallback(() => setActiveStepIndex(stepIndex => Math.max(stepIndex - 1, 0)), [])
+  const previous = useCallback(() => {
+    if (activeStepIndex > 0) {
+      setActiveStepIndex(activeStepIndex - 1)
+    }
+  }, [activeStepIndex])
+
+  useEffect(() => {
+    onStepChange?.(activeStepIndex)
+  }, [activeStepIndex, onStepChange])
```
imkesin, line 49: I don't like this use of `useEffect`.

I would prefer that this is called every time the index truly changes (inside of `next` and `previous`).

imkesin (approved):

src/components/Wizard/Wizard.tsx
sarahraines: Sure thing, I can do that!

PR #653, opened 2025-06-11
draft(customers + vendors): `SecondPartySelector`

imkesin:
## Description

This is a work in progress...


src/components/ui/TagGroup/tagGroup.scss
```diff
@@ -16,17 +16,16 @@
...
 
   gap: var(--spacing-2xs);
 
   align-items: center;
 
-  min-block-size: 2.5rem;
+  block-size: 2.25rem;
   min-inline-size: 0;
 
-  padding-block: var(--spacing-3xs);
```
imkesin, line 29: When we need to fix the height, then it makes sense to remove vertical padding

src/styles/internal_variables.scss
```diff
@@ -18,24 +18,19 @@
...
   --button-bg-active: var(--color-base-800);
-
   --button-fg-disabled: var(--color-base-500);
   --button-bg-disabled: var(--color-base-200);
-
-  --button-bg-ghost: var(--color-base-0);
   --button-fg-ghost: var(--color-base-800);
-
   --button-border-color-ghost: var(--color-base-300);
   --button-border-color-ghost-active: var(--color-base-800);
```
imkesin, line 35: I'm thinking of opening a PR to `stylelint` so that we can better control spacing between custom properties.

src/components/BankTransactions/BankTransactions.tsx
```diff
@@ -90,7 +93,9 @@ export const BankTransactions = ({
       <BankTransactionsContext.Provider value={contextData}>
         <LegacyModeProvider overrideMode={mode}>
           <BankTransactionTagVisibilityProvider showTags={showTags}>
-            <BankTransactionsContent {...props} />
+            <BankTransactionSecondPartyVisibilityProvider showSecondParty={showSecondParty}>
```
sarahraines, line 96: I'm thinking if we want to have more features like this, we may want to collapse these context providers into one `BankTransactionsFeatureVisibilityProvider`. Am I missing something about this design choice?

src/components/ui/Button/button.scss
```diff
@@ -9,19 +8,35 @@
...
   cursor: pointer;
   user-select: none;
+  font-size: var(--text-md);
   appearance: none;
   white-space: nowrap;
   vertical-align: middle;
 
+  &[data-inset] {
+    --size: 1.75rem;
+  }
```
sarahraines, line 28: Love this! is there anything else we want to do with the inset style to make the button more subtle?

src/styles/internal_variables.scss
```diff
@@ -18,24 +18,19 @@
   /*
    * Outlines
    */
-  --outline-subtle: var(--color-base-200);
-  --outline-default: var(--color-base-300);
+  --outline-subtle: var(--color-base-300);
+  --outline-default: var(--color-base-400);
```
sarahraines, line 22: I think there's a significant jump between 300 and 400 in terms of darkness, so I'm not certain about the impact this change on other places that are using `--outline-default`. 

For example, I know the new button uses to determine the outline color for its "pressed" state, and I think changing it is causing the styles to diverge between the old and new buttons.

src/features/tags/components/TagSelector.tsx
```diff
@@ -33,14 +32,6 @@ function TagSelectorLayoutGroup({ children }: PropsWithChildren) {
   )
 }
 
-function TagSelectorInputGroup({ children }: PropsWithChildren) {
```
sarahraines, line 36: Love that we're deleting this and using the base input group styles.

src/features/bankTransactions/[bankTransactionId]/secondParty/components/BankTransactionSecondPartySelector.tsx
```diff
@@ -0,0 +1,110 @@
...
+          secondPartyType: 'CUSTOMER',
+        })
+      }
+
+      if (vendor) {
+        return decodeSecondParty({
+          ...vendor,
+          secondPartyType: 'VENDOR',
+        })
+      }
```
Comment (sarahraines) [inline src/features/bankTransactions/[bankTransactionId]/secondParty/components/BankTransactionSecondPartySelector.tsx:50]: I think there's a bug here if you select a customer and then select a vendor without deleting the customer, the `selectedSecondParty` is still the customer, and the popover does not close.

src/views/BankTransactionsWithLinkedAccounts/BankTransactionsWithLinkedAccounts.tsx
```diff
@@ -22,6 +22,7 @@ export interface BankTransactionsWithLinkedAccountsProps {
   showDescriptions?: boolean
   showLedgerBalance?: boolean
   showReceiptUploads?: boolean
+  showSecondParty?: boolean
```
sarahraines, line 25: For this prop, I'm wondering if we want to do `showCounterparty` rather than second party because this aligns with the naming we have in some of our API params/responses. 

For example: https://docs.layerfi.com/api-reference/bank-transaction/bank-transaction#param-counterparty-name

src/components/BankTransactions/BankTransactions.tsx
imkesin: The answer here is a bit more involved.

Until we can tell that we know that these visibility modifiers are relied-upon in very similar places, I'm not putting them together into one context. The reflexive use of "one big context" is almost always incorrect.

(That being said... this use of `Context` isn't really an "anything" optimization. We need to use `React.memo` on the child).

src/components/ui/Button/button.scss
```diff
@@ -44,11 +59,20 @@
...
+  &[data-variant='ghost'] {
     border: 1px solid transparent;
-    background-color: var(--button-bg-ghost);
+    background-color: transparent;
     color: var(--button-fg-ghost);
 
+    &[data-pending],
+    &[data-disabled] {
+      background-color: transparent;
+    }
```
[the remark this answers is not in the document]
imkesin: They need to move in the same direction; both of them should become lower-contrast / "more invisible".

src/styles/internal_variables.scss
imkesin: Fair point - I won't step these up.

I need to spend some dedicated time trying to wrangle crazy uses of color across the codebase.

src/views/BankTransactionsWithLinkedAccounts/BankTransactionsWithLinkedAccounts.tsx
imkesin: I need to rename this into `CustomerVendor`

src/components/ui/Modal/ModalSlots.tsx
```diff
@@ -15,7 +15,6 @@ export function ModalContextBar({ onClose }: ModalContextBarProps) {
     <div className={MODAL_CONTEXT_BAR_CLASS_NAME}>
       <Button
         icon
-        size='sm'
```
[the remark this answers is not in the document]
imkesin: I've addressed this one in a few spots.

PR #657, opened 2025-06-12
fix(bank-transaction): remove animation that renders invisible items

imkesin:
## Description

When we were transitioning the height of a bank transaction row, we were not actually hiding the elements from the UI. This resulting in some strange experiences when trying to navigate with a keyboard.

### DIFF
--- a/src/components/ExpandedBankTransactionRow/ExpandedBankTransactionRow.tsx
+++ b/src/components/ExpandedBankTransactionRow/ExpandedBankTransactionRow.tsx
@@ -2,10 +2,7 @@ import {
   forwardRef,
   useImperativeHandle,
   useState,
-  useCallback,
-  useEffect,
   useRef,
-  TransitionEvent,
 } from 'react'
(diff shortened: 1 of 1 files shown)


src/components/ExpandedBankTransactionRow/ExpandedBankTransactionRow.tsx
```diff
@@ -383,17 +347,14 @@ const ExpandedBankTransactionRow = forwardRef<SaveHandle, Props>(
...
     return (
       <span
         className={`${className} ${className}--${
           isOpen ? 'expanded' : 'collapsed'
         }`}
-        style={{ height }}
-        onTransitionEnd={handleTransitionEnd}
       >
-        {shouldHide
+        {!isOpen
```
sarahraines, line 357: NIT: might be more intuitive to un-invert this to have the condition be `isOpen` and then the "true" part of the ternary be the component.

sarahraines (approved):
🔪 animations begone! 🙈

PR #661, opened 2025-06-13
fix: incorporate design feedback from Justin

sarahraines:
## Description
A batch of cosmetic changes requested by Justin to CSV upload, including
1. Restyle custom accounts form and do not show rest of modal contents when custom account form is showing.
2. Only show additional account data (institution name, type) in the select account dropdown menu, not the input.
3. Change the appearance of some CTAs.
4. Zhuzh content design.

### DIFF
--- a/src/components/CustomAccountForm/customAccountForm.scss
+++ b/src/components/CustomAccountForm/customAccountForm.scss
@@ -1,5 +1,12 @@
 .Layer__custom-account-form {
-  .Layer__custom-account-form__input {
-    width: 100%;
+  .Layer__custom-account-form__field {
+    >.Layer__input-tooltip {
+      display: flex;
+      flex: 1 1;
+
+      >.Layer__custom-account-form__input {
+        flex: 1 1;
+      }
+    }
   }
 }
--- a/src/components/CustomAccountForm/CustomAccountForm.tsx
+++ b/src/components/CustomAccountForm/CustomAccountForm.tsx
@@ -47,7 +47,7 @@ export const CustomAccountForm = ({ initialAccountName, onCancel, onSuccess }: C
         >
           {field => (
             <>
-              <InputGroup name='account_name' label='Account name'>
+              <InputGroup name='account_name' label='Account name' className='Layer__custom-account-form__field'>
                 <Input
                   className='Layer__custom-account-form__input'
                   name='account_name'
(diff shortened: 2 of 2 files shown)


src/components/CsvUpload/csvUpload.scss
```diff
@@ -41,10 +41,10 @@
 
 .Layer__csv-upload__file-row {
   height: 2rem;
-  border-radius: var(--border-radius-3xs);
   padding: var(--spacing-xs) var(--spacing-sm);
+  border-radius: var(--border-radius-3xs);
+  border: 1px solid var(--color-base-300);
```
sarahraines, line 46: This is the relevant change

src/components/CustomAccountForm/customAccountForm.scss
```diff
@@ -0,0 +1,5 @@
+.Layer__custom-account-form {
+  .Layer__custom-account-form__input {
+    width: 100%;
+  }
+}
```
imkesin, line 5: This makes me very suspicious.

We should almost never need `width: 100%` - it's a bad practice. I'm pretty sure you mean that the parent of the input should be `display: grid`.

src/components/UploadTransactions/UploadTransactionsUploadCsvStep.tsx
```diff
@@ -136,7 +141,7 @@ export function UploadTransactionsUploadCsvStep(
           className={inputClassName}
         />
       </VStack>
-      {selectedAccount && selectedAccount.value === 'new_account' && (
+      {isCreatingNewAccount && (
```
imkesin, line 144: The stuff below looks like it should be a ternary.

src/components/UploadTransactions/uploadTransactions.scss
```diff
@@ -10,7 +10,6 @@
   .Layer__upload-transactions__select-account-name-input {
     .Layer__select__control {
       width: 24rem;
-      min-height: 44px;
```
imkesin, line 13: 🍾 🍾

imkesin (approved):

src/components/CustomAccountForm/customAccountForm.scss
sarahraines: Talked offline - the tooltip is cursed. Did a slightly better rev with flex, but ultimately a problem for another PR.

PR #668, opened 2025-06-18
feat: add hooks for interacting with tags on the ledger

imkesin:
## Description

These changes are intended to set-up future UI work where ledger entries can be tagged.

### DIFF
--- a/src/components/JournalTable/JournalTableWithPanel.tsx
+++ b/src/components/JournalTable/JournalTableWithPanel.tsx
@@ -54,13 +54,19 @@ export const JournalTableWithPanel = ({
     addEntry,
   } = useContext(JournalContext)
 
-  const data = useMemo(() => {
-    const firstPageIndex = (currentPage - 1) * pageSize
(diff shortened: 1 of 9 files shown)


src/components/JournalTable/JournalTableWithPanel.tsx
```diff
@@ -54,13 +54,19 @@ export const JournalTableWithPanel = ({
     addEntry,
   } = useContext(JournalContext)
 
-  const data = useMemo(() => {
-    const firstPageIndex = (currentPage - 1) * pageSize
-    const lastPageIndex = firstPageIndex + pageSize
-    return rawData
-      ?.sort((a, b) => Date.parse(b.entry_at) - Date.parse(a.entry_at))
```
imkesin, line 61: I removed this `sort` because the sort should now be enforced by the sever.

src/types/journal.ts
imkesin, line 1: The types need some work, but that's a major overhaul.

PR #671, opened 2025-06-18
fix: delete outlined button special hover state

sarahraines:
## Description
Fixes [LAY-1488](https://linear.app/layerfi/issue/LAY-1488/fix-hover-state-outline-for-txn-upload-button)

I think the ghost and outlined buttons should have the same hover state, even if it is more subtle for the outlined button.

### DIFF
--- a/src/components/ui/Button/button.scss
+++ b/src/components/ui/Button/button.scss
@@ -85,10 +85,6 @@
       color: var(--button-fg-disabled);
     }
 
-    &[data-hovered]:not([data-pressed]) {
-      border-color: var(--button-border-color-ghost-active);
-    }
-
     &[data-pressed]:not([data-focus-visible]) {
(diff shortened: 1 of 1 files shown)


src/components/ui/Button/button.scss
```diff
@@ -1,4 +1,5 @@
 .Layer__UI__Button {
+  all: unset;
```
sarahraines, line 2: This is load bearing for our solid button styles - it's not ideal, but let's leave here for now until we can make sure that we've got all states working.

imkesin: What do they do without `all: unset`?

I originally had this to make stuff super clear... but then I was realizing it might prevent me from inheriting some good default behaviors.

imkesin (approved):

PR #674, opened 2025-06-19
fix: when a value is selected in a `ComboBox`, still show all items

imkesin:
## Description

Once a user has made a ComboBox selection - e.g. they picked a customer - the dropdown should not be filtered to only show their choice. We should present the unfiltered list to them.

### Note

I can't get the special behavior of typing over a result to work without playing bug whack-a-mole. Everything that I do results in an issue in another area.


imkesin:
This isn't going to be the right solution - we need to move to another library entirely

PR #683, opened 2025-06-25
fix: increase (relative) width of categorization input

imkesin:
## Description

When there is a "Match" badge within the categorization combobox, it crowds-out the meaningful info. The short-term solution is to change the ratios of the table.


imkesin (commented):
This is more "hack" than engineering work.

We clearly need to move toward a more sustainable (and understandable solution).

PR #686, opened 2025-06-25
fix: primary and accent color contrast

imkesin:
## Description

This PR introduces 4 undefined CSS variables that must be explicitly set by the platform:

```
--bg-brand-primary
--fg-brand-primary
 
--bg-brand-accent
--fg-brand-accent
```

When these variables are set, they will override the use of `color-primary` (think: "dark") and `color-accent` (think: "light") in two (2) key areas:
- The hovered state of some solid buttons
- The default badge color


sarahraines (approved):
LGTM!

PR #693, opened 2025-06-27
fix: point bank transaction search to use the `q` parameter

imkesin:
## Description

By using the `q` query parameter, we can get filtering by amount.


AndrewAcomb (approved):
Looks great

PR #676, opened 2025-06-20
fix: selector width matches parent, allow flipping and scrolling

sarahraines:
## Description
Fixes LAY-1481 and some other goodies (comments inline).

PR #702, opened 2025-07-03
fix: make `TagSelector` look more like `CustomerVendorSelector`

imkesin:
## Description

I'm now pointing the `TagSelector` component at my improved `ComboBox` implementation (backed by `react-select`).


src/features/customerVendor/components/CustomerVendorSelector.tsx
```diff
@@ -235,22 +239,40 @@ export function CustomerVendorSelector({
...
+
+        onInputValueChange={handleInputChange}
+
+        placeholder={placeholder}
+        slots={{ EmptyMessage, ErrorMessage }}
+
+        isDisabled={isReadOnly || shouldDisableComboBox}
+        isError={isError}
+        isLoading={isLoadingWithoutFallback}
+      />
```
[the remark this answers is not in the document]
imkesin: Great catch. Thank you, Greptile.

src/components/ui/ComboBox/ComboBox.tsx
```diff
@@ -216,19 +231,26 @@ export function ComboBox<T extends ComboBoxOption>({
...
-  const inputId = useId()
+  const internalInputId = useId()
+  const effectiveInputId = inputId ?? internalInputId
 
   const CustomGroupHeadingRef = useRef(buildCustomGroupHeading())
-  const CustomComboBoxOptionRef = useRef(buildCustomComboBoxOption())
+  const CustomComboBoxOption = useMemo(
+    () => (buildCustomComboBoxOption({ displayDisabledAsSelected })),
+    [displayDisabledAsSelected],
+  )
```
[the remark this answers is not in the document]
imkesin: If somehow this value changes, then it needs to be switched too.

src/components/ui/ComboBox/ComboBox.tsx
```diff
@@ -70,18 +71,29 @@ function buildCustomGroupHeading() {
...
   }: OptionProps<T, false, GroupBase<T>>) {
+    const { isSelected, isFocused, isDisabled } = restProps
+
+    const effectiveIsSelected = isSelected || (displayDisabledAsSelected && isDisabled)
+
     return (
       <components.Option
         {...restProps}
         className={classNames(
           COMBO_BOX_CLASS_NAMES.OPTION,
```
sarahraines, line 93: The disabled/already selected option still has a hover state - let's only show that hover state when the user is able to select the option.
Screenshot 2025-07-07 at 10 17 14 AM

src/features/tags/components/TagSelector.tsx
```diff
@@ -177,65 +174,91 @@ function LabeledTagSelectorContainer({
...
+  const inputId = useId()
 
-  type TItemDerived = Exclude<typeof data, undefined> extends ReadonlyArray<infer TItem>
-    ? TItem
-    : never
+  const EmptyMessage = useMemo(
+    () => (
+      <Span
+        variant='subtle'
+        nonAria
```
sarahraines, line 243: You'll have to remind me - what was the reason we're choosing to use base `span` here rather than `ReactAriaText`?

src/features/customerVendor/components/CustomerVendorSelector.tsx
```diff
@@ -235,22 +239,41 @@ export function CustomerVendorSelector({
...
-      label='Customer or Vendor'
-      placeholder={placeholder}
-      slots={{ EmptyMessage, ErrorMessage }}
-
-      isDisabled={isReadOnly || shouldDisableComboBox}
-      isError={isError}
-      isLoading={isLoadingWithoutFallback}
-      isMutating={isMutating}
-    />
+    <VStack gap='3xs'>
```
sarahraines, line 242: NIT: Maybe not for this PR, but do you think we should we generalize this to some sort of `LabeledComboBox` component that includes both the input and the label? Or maybe we want a general `LabeledInput` component for all sorts of inputs, including plain text inputs, textareas, and comboboxes. 

Thinking towards a future where we want visual consistency across labeled inputs and to easily be able to spin up forms.

sarahraines (commented):
Looks great! Found one small thing to fix (disabled hover styles).

Was also wondering if we could fold these stylistic changes from [my old PR](https://github.com/Layer-Fi/layer-react/pull/676/commits/b0c2c31d22ed88f30771d8797ec63a4232a4b884), specifically to `inputGroup.scss`, as well as apply them to the new combobox. 

The goal would be to get the input to look like our button styles when it's active:
Screenshot 2025-07-07 at 10 36 48 AM

src/features/tags/components/TagSelector.tsx
imkesin: Good point - I can get rid of this now.

It's because `react-aria` is looking for text elements in certain positions.

src/components/ui/ComboBox/comboBox.scss
```diff
@@ -132,9 +146,8 @@
   &--focused {
     background-color: var(--color-base-50);
 
-    &--selected,
     &:hover {
-      border: 2px solid var(--border-color);
+      border-color: var(--border-color);
```
sarahraines, line 150: NIT: I kinda think `--border-color` should be `--color-basee-800`

PR #706, opened 2025-07-12
draft: add `(CustomerVendor|Tag)Selector` on journal entry

imkesin:
## Description

This is a start on the work needed to allow tagging and setting a customer or vendor on journal entries.

### DIFF
--- a/src/components/JournalEntryDetails/JournalEntryDetails.tsx
+++ b/src/components/JournalEntryDetails/JournalEntryDetails.tsx
@@ -17,6 +17,9 @@ import { Header, HeaderCol, HeaderRow } from '../Header'
 import { SourceDetailView } from '../LedgerAccountEntryDetails/LedgerAccountEntryDetails'
 import { Table, TableBody, TableCell, TableHead, TableRow } from '../Table'
 import { Heading, HeadingSize } from '../Typography'
+import { JournalEntryTagSelector } from '../../features/ledger/entries/[ledgerEntryId]/tags/components/JournalEntryTagSelector'
+import { VStack } from '../ui/Stack/Stack'
+import { JournalEntryCustomerVendorSelector } from '../../features/ledger/entries/[ledgerEntryId]/customerVendor/components/JournalEntryCustomerVendorSelector'
 
 export const JournalEntryDetails = () => {
   const {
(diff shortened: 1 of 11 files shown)


src/features/ledger/entries/api/useListLedgerEntries.ts
imkesin, line 1: The changes in this file are required because we converted this to an infinite SWR in a previous PR.

src/features/ledger/entries/[ledgerEntryId]/metadata/api/useSetMetadataOnLedgerEntry.ts
Comment (imkesin) [inline src/features/ledger/entries/[ledgerEntryId]/metadata/api/useSetMetadataOnLedgerEntry.ts:1]: This file is not finished because we don't have a PATCH endpoint for ledger entry metadata.

src/components/ui/ComboBox/comboBox.scss
```diff
@@ -95,6 +95,14 @@
...
 }
 
+.Layer__ComboBoxMenuPortal {
+  /*
+   * We are limited by the default behavior of `react-select`, which imposes a `z-index: 1`
+   * on the menu portal and the portal by default.
+   */
+  z-index: 100 !important;
+}
+
```
imkesin, line 105: This type of change should require eyes from @sarahraines... it's bad practice, but we have limited options.

An alternate solution would involve eliminating the `z-index` on our panel components entirely.

src/components/JournalEntryDetails/JournalEntryDetails.tsx
```diff
@@ -118,115 +125,142 @@ export const JournalEntryDetails = () => {
...
+        : null}
+
+      {entry
+        ? (
+          /**
+           * @see {BankTransactionFormFields}
+           *
+           * This could be extracted out into a component, especially if it needs to
+           * be conditionally rendered based on some props.
+           */
```
imkesin, line 252: If we don't want these fields to flicker (on the very first load), then we may want to do a preload higher-up in the DOM.

src/components/JournalEntryDetails/JournalEntryDetails.tsx
imkesin, line 1: A good chunk of changes in this file are purely auto-formatting. I've tried to split this out into a separate commit.

PR #710, opened 2025-07-22
@container change for vertical stacking

doneel:
### Before
image

### After
image

### DIFF
--- a/src/styles/actionable_row.scss
+++ b/src/styles/actionable_row.scss
@@ -7,12 +7,28 @@
   border-radius: var(--XS, 8px);
   background: var(--color-base-0);
   box-shadow: 0px 0px 0px 1px var(--color-base-300);
+  container-type: inline-size;
+
+  /* Stack vertically when container is too narrow */
+  @container (max-width: 400px) {
+    flex-direction: column;
+    align-items: stretch;
+
+    .Layer__actionable_row__main {
+      width: 100%;
+    }
+
+    .Layer__actionable_row__action {
+      justify-content: flex-end;
+    }
+  }
 }
 
 .Layer__actionable_row__main {
   display: flex;
   align-items: center;
   gap: var(--spacing-md);
+  flex: 1;
 }
 
 .Layer__actionable_row__icon {
@@ -24,6 +40,12 @@
   align-items: center;
   justify-content: center;
   box-shadow: 0px 0px 3px 0px var(--base-transparent-4) inset;
+}
+
+.Layer__actionable_row__action {
+  display: flex;
+  align-items: center;
+  flex-shrink: 0; /* Prevent action area from shrinking, maintaining button sizes */
 }
 
 .Layer__actionable_row__description {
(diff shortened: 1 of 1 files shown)


src/styles/actionable_row.scss
```diff
@@ -7,12 +7,28 @@
...
+  container-type: inline-size;
+
+  /* Stack vertically when container is too narrow */
+  @container (max-width: 400px) {
+    flex-direction: column;
+    align-items: stretch;
+
+    .Layer__actionable_row__main {
+      width: 100%;
+    }
```
sarahraines, line 19: Poked around with removing styles and I _think_ this one isn't necessary, but lmk if there's something I'm missing here!

src/styles/actionable_row.scss
```diff
@@ -7,12 +7,28 @@
   border-radius: var(--XS, 8px);
   background: var(--color-base-0);
   box-shadow: 0px 0px 0px 1px var(--color-base-300);
+  container-type: inline-size;
+
+  /* Stack vertically when container is too narrow */
+  @container (max-width: 400px) {
```
sarahraines, line 13:
```suggestion
  @container (max-width: 30rem) {
```

Let's stack a bit larger so the text doesn't go over two lines (30 rem === 480px) and also prefer to use rem here rather than px values to scale with the user's text size (if they change a11y settings).

src/styles/actionable_row.scss
```diff
@@ -26,6 +42,12 @@
   box-shadow: 0px 0px 3px 0px var(--base-transparent-4) inset;
 }
 
+.Layer__actionable_row__action {
+  display: flex;
+  align-items: center;
+  flex-shrink: 0; /* Prevent action area from shrinking, maintaining button sizes */
```
sarahraines, line 48: I'm pretty certain the button won't shrink here, since the size is based on the interior content. These lines may not be necessary.

src/styles/actionable_row.scss
```diff
@@ -7,12 +7,28 @@
...
+      justify-content: flex-end;
+    }
+  }
 }
 
 .Layer__actionable_row__main {
   display: flex;
   align-items: center;
   gap: var(--spacing-md);
+  flex: 1;
```
sarahraines, line 31: Also not certain this is necessary.

sarahraines (approved):
Looks great! Some nits - I'm not certain all the added CSS is required, but I could be missing some buggy state that you saw.

src/styles/actionable_row.scss
doneel: I think you're right, removing.

src/styles/actionable_row.scss
doneel: I think I need this here - it's not the button itself shrinking but the space it takes up which makes it swim left instead of being right aligned: 

Image

src/styles/actionable_row.scss
doneel: You're right - I needed this before the flex settings below but now this is removable

PR #713, opened 2025-07-23
feat: show plaid link flow error

sarahraines:
## Description
This PR shows an error when Plaid is unable to link a bank rather than failing silently.

As an error state In the `LinkAccounts` component
Screenshot 2025-07-22 at 7 11 14 PM

As a dismissable modal in the `LinkedAccounts` component
Screenshot 2025-07-22 at 7 10 06 PM

### DIFF
--- a/src/components/LinkedAccounts/PlaidLinkError/PlaidLinkError.tsx
+++ b/src/components/LinkedAccounts/PlaidLinkError/PlaidLinkError.tsx
@@ -42,7 +42,7 @@ export function PlaidLinkErrorContent({ onComplete, onPlaidLink }: PlaidLinkErro
           Link another bank
         </DeprecatedButton>
         <DeprecatedButton
-          onClick={() => { onComplete() }}
+          onClick={onComplete}
           rightIcon={<ChevronRight />}
           className='Layer__plaid-link-error__button-row-item'
         >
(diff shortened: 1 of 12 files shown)


src/components/LinkedAccounts/PlaidLinkError/plaidLinkError.scss
```diff
@@ -0,0 +1,19 @@
+.Layer__plaid-link-error__content {
+  container-type: inline-size;
+
+  .Layer__plaid-link-error__data-state {
+    .Layer__data-state__description {
+      margin-bottom: 0;
+    }
+  }
+
+  @container (max-width: 26.25rem) {
```
doneel, line 10: Where does this number come from? Asking to see if it's a repeatable constant we should re-use as a var.

src/components/LinkedAccounts/PlaidLinkError/PlaidLinkError.tsx
```diff
@@ -0,0 +1,76 @@
+import { useCallback, useContext } from 'react'
```
doneel, line 1: I think this is inconsequential, but why should this `PlaidLinkError` component be nested under `LinkedAccounts`? We're reusing it for multiple other components.

doneel (approved):

PR #724, opened 2025-07-29
feat: create new checkbox field and use in invoices form

sarahraines:
## Description
This PR creates a new `FormCheckboxField` and uses it in the invoices form.

Screenshot 2025-07-29 at 10 34 47 AM

### DIFF
--- a/src/styles/internal_variables.scss
+++ b/src/styles/internal_variables.scss
@@ -35,20 +35,4 @@
 
   --button-border-color-ghost: var(--color-base-300);
   --button-border-color-ghost-active: var(--color-base-800);
-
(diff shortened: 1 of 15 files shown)


src/styles/internal_variables.scss
```diff
@@ -35,20 +35,4 @@
 
   --button-border-color-ghost: var(--color-base-300);
   --button-border-color-ghost-active: var(--color-base-800);
-
-  /*
-   * Checkbox
-   */
```
sarahraines, line 41: Moved these styles into the checkbox css styles itself because they're not needed elsewhere.

PR #734, opened 2025-08-05
feat: allow setting reference num and external id in csv upload

AndrewAcomb:
## Description
Per a customer request, provide a way to set reference numbers (e.g., check numbers) while uploading bank transactions

## How this has been tested?
Screenshot 2025-08-05 at 3 10 09 PM

### DIFF
--- a/src/components/CsvUpload/ValidateCsvTable.tsx
+++ b/src/components/CsvUpload/ValidateCsvTable.tsx
@@ -52,21 +52,9 @@ export function ValidateCsvTable<T extends { [K in keyof T]: string | number | n
         cell: (info: CellContext<PreviewRow<T>, unknown>) => {
           const field = info.row.original[key]
 
-          // Handle null/undefined fields for optional columns
-          if (!field) {
-            return (
-              <span className={classNames(
-                'Layer__table-cell-content',
-                `${CSS_PREFIX}__cell-content`,
-              )}
-              >
-                {/* Empty cell for null values */}
-              </span>
-            )
-          }
-
-          let value: string | number | null | undefined = field.raw
-          if (field.is_valid) {
+          let value: string | number | null | undefined = field?.raw
+          const isValid = field && field.is_valid
+          if (isValid) {
             const formatter = formatters?.[key]
             value = formatter ? formatter(field.parsed as T[keyof T]) : field.parsed
           }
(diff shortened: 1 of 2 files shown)


src/components/CsvUpload/ValidateCsvTable.tsx
```diff
@@ -47,7 +52,20 @@ export function ValidateCsvTable<T extends { [K in keyof T]: string | number }>(
...
+                `${CSS_PREFIX}__cell-content`,
+              )}
+              >
+                {/* Empty cell for null values */}
+              </span>
+            )
+          }
+
+          let value: string | number | null | undefined = field.raw
           if (field.is_valid) {
```
sarahraines, line 69: If value is falsy (undefined, null, etc.) then nothing will get rendered inside the cell anyway. This should work:
```suggestion
          let value: string | number | null | undefined = field.raw
          const isValid = field && field.is_valid
          if (isValid) {
```

src/components/UploadTransactions/UploadTransactionsValidateCsvStep.tsx
```diff
@@ -25,6 +27,28 @@ const formatters = {
   amount: (parsed: number) => convertCentsToCurrency(parsed) ?? '',
 }
 
+const generateDynamicHeaders = (transactionsPreview: PreviewCsv<CustomAccountTransactionRow>) => {
```
sarahraines, line 30: For ordering, let's do the following:
1. Row number
2. External ID
3. Reference No.
4. Date 
5. Description
6. Amount

src/components/UploadTransactions/UploadTransactionsValidateCsvStep.tsx
```diff
@@ -25,6 +27,28 @@ const formatters = {
...
+  const dynamicHeaders = { ...templateHeaders }
```
sarahraines, line 31: The columns get ordered in the order the headers are defined in templateHeaders. To get the headers in the right order, we can either: 

(1) Include all potential headers in `templateHeaders` to allow for correct ordering, and then conditionally subtract unnecessary headers.

OR

(2) Generate dynamic headers as follows:
```
const dynamicHeaders = {
  ...(hasExternalId && { external_id: 'External ID' ),
  ...(hasReferenceNumber && { reference_number: 'Reference No.' )
  ...templateHeaders
}
```

src/components/UploadTransactions/UploadTransactionsValidateCsvStep.tsx
AndrewAcomb: ✅

sarahraines (approved):
LGTM!

PR #735, opened 2025-08-06
Reconnect plaid accounts with new credentials if necessary

AndrewAcomb:
## Description
Either repair an existing a connection or create a new one depending on whether the existing connection is on the platform's current Plaid credentials.


src/components/LinkedAccounts/LinkedAccountItemThumb.tsx
```diff
@@ -82,6 +82,7 @@ export const LinkedAccountItemThumb = ({
               // An account is "broken" when its connection is broken
               void repairConnection(
```
sarahraines, line 83: As (mostly) a matter of style, I'd prefer if we didn't change the functionality of `repairConnection` here and instead called `addConnection` instead if we needed the new credentials set.

This will help prevent drift if the two implementations ever change.

sarahraines (approved):

PR #744, opened 2025-08-10
Added tooltip message about sub accounts for account deletion

layer-shyam:
## Description
Added tooltip message about sub accounts for account deletion
## Changes
Added tooltip message to detail that you can not delete accounts with sub accounts

### DIFF
--- a/src/components/ChartOfAccountsTable/ChartOfAccountsTable.tsx
+++ b/src/components/ChartOfAccountsTable/ChartOfAccountsTable.tsx
@@ -140,6 +140,9 @@ export const ChartOfAccountsTableContent = ({
     if (account.balance !== 0) {
       return 'This account cannot be deleted because it has ledger entries'
     }
+    if (account.sub_accounts.length > 0) {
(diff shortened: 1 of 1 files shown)


AndrewAcomb (approved):

src/components/ChartOfAccountsTable/ChartOfAccountsTable.tsx
```diff
@@ -140,6 +140,9 @@ export const ChartOfAccountsTableContent = ({
     if (account.balance !== 0) {
       return 'This account cannot be deleted because it has ledger entries'
     }
+    if (account.sub_accounts.length > 0) {
+      return 'This account cannot be deleted because it has sub-accounts'
+    }
```
sarahraines, line 145: I believe this one should go before the `account.balance` check, as the balance can be >0 even if there are no ledger entries in the account (because there are ledger entries in the sub account).

layer-shyam: Resolved

sarahraines (approved):

PR #745, opened 2025-08-10
Feat: Link to things outside Layer!

AndrewAcomb:
Fixes: LAY-1578

## Description
This PR allows you to specify functions to generate links to page in the rest of your experience. For example you can link from a suggested match for an invoice payment to the invoice the payment is for.

## Changes
Adds 2 props- `convertMatchDetailsToSourceLink` for BankTransactions / BankTransactionsWithLinkedAccounts and `convertLedgerEntrySourceToSourceLink` for GeneralLedger / ChartOfAccounts / Journal. 

That said, I don't think it would be crazy to just put both contexts just under the LayerContext, since we will later need to add convertLedgerEntrySourceToSourceLink under Reports and Overview (for example, for P&L details).

## How this has been tested?

Desktop expanded view
Screenshot 2025-08-10 at 8 18 27 PM

Desktop suggestion
Screenshot 2025-08-10 at 8 39 37 PM

Mobile suggestion
Screenshot 2025-08-10 at 10 01 17 PM

Mobile details
Screenshot 2025-08-10 at 10 01 22 PM

Chart of Accounts ledger entry details
Screenshot 2025-08-10 at 11 05 21 PM

Journal ledger entry details
Screenshot 2025-08-10 at 11 05 31 PM


src/components/MatchForm/MatchFormMobile.tsx
```diff
@@ -55,6 +58,20 @@ export const MatchFormMobile = ({
...
+                      target={sourceLink.target}
+                      onClick={e => e.stopPropagation()}
+                      className={`${classNamePrefix}__match-item__link`}
+                    >
+                      <Text size={TextSize.sm} as='span'>
+                        {sourceLink.text}
+                      </Text>
+                    </a>
+                  )
+                  : <div></div>}
```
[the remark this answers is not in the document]
AndrewAcomb: This is an intentional hack since we do space-between in order to align the link on the left and the date on the right, but I suppose I should find another way? I would need to adjust the CSS depending on whether there is a link for the suggested match.

src/views/GeneralLedger/GeneralLedger.tsx
```diff
@@ -25,13 +27,15 @@ export interface GeneralLedgerProps {
   showTitle?: boolean
   stringOverrides?: GeneralLedgerStringOverrides
   chartOfAccountsOptions?: ChartOfAccountsOptions
+  convertLedgerEntrySourceToSourceLink?: (source: LedgerEntrySource) => SourceLink | undefined
```
sarahraines, line 30: Thinking about naming here - I want to specify 1) the shape of the function that needs to be passed, and 2) the result that it will have (i.e., showing a link). Something like: 
```suggestion
  withLedgerEntrySourceLink?: (source: LedgerEntrySource) => SourceLink | undefined
```

src/types/ledger_accounts.ts
```diff
@@ -42,66 +42,185 @@ export interface LedgerAccountLineItem {
...
+
+export interface ApiVendorData {
+  vendor_external_id?: string
+  vendor_description: string
+}
+
+export interface ApiCustomerData {
+  customer_description: string
+  recipient_name?: string
+}
```
sarahraines, line 60: More schemas here too please to use as subschemas for the below!

src/types.ts
```diff
@@ -16,7 +16,23 @@ export {
...
   DisplayState,
+  SuggestedMatch,
+  BankTransactionMatch,
 } from './types/bank_transactions'
+export {
+  MatchDetails,
+  ManualJournalEntryMatchDetails,
+  RefundPaymentMatchDetails,
+  VendorRefundPaymentMatchDetails,
+  // InvoicePaymentMatchDetails,
```
sarahraines, line 27: This guy got left out 😭

src/components/MatchForm/MatchFormMobile.tsx
sarahraines, line 74: You can do some CSS, but this should be almost same affect as yours, and looks slightly better codewise.
```suggestion
              <HStack>
                {sourceLink && (
                    <a
                      href={sourceLink.href}
                      target={sourceLink.target}
                      onClick={e => e.stopPropagation()}
                      className={`${classNamePrefix}__match-item__link`}
                    >
                      <Text size={TextSize.sm} as='span'>
                        {sourceLink.text}
                      </Text>
                    </a>
                  )}
                </HStack>
```

src/components/LedgerAccountEntryDetails/LedgerAccountEntryDetails.tsx
```diff
@@ -146,12 +137,12 @@ export const SourceDetailView = ({
         </>
       )
     }
-    case 'Refund_Ledger_Entry_Source': {
-      const refundSource = source as RefundLedgerEntrySource
+    case 'Refund_Allocation_Ledger_Entry_Source': {
+      const refundSource = source as CustomerRefundAllocationLedgerEntrySource
```
sarahraines, line 141: If we use schemas, I _believe_ we don't need to do any type assertion here.
```
export const SourceDetailView = ({
  source,
  stringOverrides,
}: {
  source: LedgerEntrySource
  stringOverrides?: SourceDetailStringOverrides
}) => {
switch (source.type) {
    case 'Customer_Refund_Payment_Ledger_Entry_Source': {
      // TS knows this is CustomerRefundPaymentLedgerEntrySource
      return source.customer_id // ✅ typed
    }
    ...
  }
}
```

src/components/MatchForm/MatchFormMobile.tsx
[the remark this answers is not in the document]
sarahraines: https://github.com/Layer-Fi/layer-react/pull/745/files#r2274514498

src/styles/bank_transactions.scss
```diff
@@ -953,12 +953,25 @@
   }
 
   .Layer__expanded-bank-transaction-row__match-table__amount {
+    width: 100px;
     padding: var(--spacing-xs) var(--spacing-sm);
-    text-align: right;
   }
 
+  .Layer__expanded-bank-transaction-row__match-table__link {
+    width: 80px;
```
sarahraines, line 961: CSS kickflip - We can specify exactly how many characters we want shown if we use the `ch` modifier, e.g.,
```
width: 12ch;
```
would show 12 characters. 😎

src/components/MatchForm/MatchForm.tsx
```diff
@@ -111,21 +118,30 @@ export const MatchForm = ({
               $
               {formatMoney(suggestedMatch.details.amount)}
             </div>
+            { inAppLink && (
+              <div className={`${classNamePrefix}__match-table__link`}>
+                {inAppLink}
+              </div>
```
doneel, line 124: Image
I would like to make an effort to gracefully handle our customers putting in different sized elements than we expect. Here we have graceful vertical expansion, but not horizontal. 

It's not pleasant, but managed [made tweaks here](https://github.com/Layer-Fi/layer-react/compare/andrew/links...daniel/css-enablement-example?expand=1) as an example.

Image

src/components/BankTransactions/BankTransactions.tsx
```diff
@@ -72,6 +74,7 @@ export interface BankTransactionsProps {
   filters?: BankTransactionFilters
   hideHeader?: boolean
   stringOverrides?: BankTransactionsStringOverrides
+  convertMatchDetailsToInAppLink?: (details: MatchDetailsType) => ReactNode
```
doneel, line 77: I have a strong, strong desire not to expose the `MatchDetailsType` (or analogous types) here. I would _love_ to settle on a standard set of fields that are provided  (I'd suggest entity type, id, external id, reference number, `metadata`) and then always provide that. And if we don't have that in these types, add it in on our side so we can have that standard experience everywhere.

src/components/MatchForm/MatchForm.tsx
AndrewAcomb: Fixed!

PR #748, opened 2025-08-12
Daniel/lay 1557 reusable table virtualization

doneel:


src/components/VirtualizedDataTable/VirtualizedDataTable.tsx
```diff
@@ -0,0 +1,185 @@
+import { useRef } from 'react'
+import classNames from 'classnames'
+import { useVirtualizer, VirtualItem, Virtualizer } from '@tanstack/react-virtual'
+import { Loader } from '../Loader/Loader'
+import type { ColumnConfig, Column } from '../DataTable/DataTable'
+
+const DEFAULT_ROW_HEIGHT = 52
+const DEFAULT_OVERSCAN = 5
+const DEFAULT_TABLE_HEIGHT = 500 // Shows ~9-10 rows at default row height
```
sarahraines, line 9: For `DEFAULT_TABLE_HEIGHT`, to get exactly 10 rows, you'll want:
```suggestion
const DEFAULT_NUM_ROWS = 10
const DEFAULT_TABLE_HEIGHT = (DEFAULT_ROW_HEIGHT * DEFAULT_NUM_ROWS) + HEADER_HEIGHT - 1
```

src/components/VirtualizedDataTable/VirtualizedDataTable.tsx
```diff
@@ -0,0 +1,185 @@
...
+      ref={containerRef}
+      style={{ height }}
+      role='table'
+      aria-label={ariaLabel}
+    >
+      <table className={classNames('Layer__table', `Layer__UI__Table__${componentName}`)}>
+        <thead className={`${CSS_PREFIX}__header`}>
+          <tr className={`${CSS_PREFIX}__header-row`} style={{ height: HEADER_HEIGHT }}>
+            {columns.map(col => (
+              <th
```
sarahraines, line 102: The state of the art is to replace HTML table elements with _shiny, new_ ✨ React Aria based table components from the `/components/ui` directory, which will ensure correct styling and accessibility - no need to pass the `Layer__table` prefixed class names anymore after that.

The only difference is that they may require slight modification to take a `style` prop.

https://github.com/Layer-Fi/layer-react/blob/60a02981f9f7a5dac089cd8ab0565e74fbc40b87/src/components/ui/Table/Table.tsx#L158-L165

src/components/VirtualizedDataTable/VirtualizedDataTable.tsx
```diff
@@ -0,0 +1,185 @@
...
+            ))}
+          </tr>
+        </thead>
+        <tbody>
+          <tr
+            className={`${CSS_PREFIX}__spacer`}
+            style={{ height: totalSize }}
+          />
+          {virtualItems.map((virtualRow) => {
+            const row = data![virtualRow.index]
```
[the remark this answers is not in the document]
sarahraines: I think we may want to get `useReactTable` in here to manage the rows - this should guarantee that we have rows rather than asserting `data` to be nonnull, and also helps us not have to pass columns down to the row component.

 With `useReactTable`, we can also apply any arbitrary sorting/filtering/hiding columns, etc. to the table, which will make this component more powerful.

```
  const table = useReactTable({
    data,
    columns,
    state,
    getCoreRowModel: getCoreRowModel(),
    getSortedRowModel: getSortedRowModel(),
  })

  const { rows } = table.getRowModel()
```

And then:
```

```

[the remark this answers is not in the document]
doneel: Done!

src/components/VirtualizedDataTable/VirtualizedDataTable.tsx
```diff
@@ -0,0 +1,222 @@
...
+          {virtualItems.map((virtualRow) => {
+            const rowData = processedData[virtualRow.index]
+            return (
+              <VirtualizedDataTableRow
+                key={rowData.id}
+                rowData={rowData}
+                virtualRow={virtualRow}
+                rowVirtualizer={rowVirtualizer}
+                componentName={componentName}
+                cssPrefix={CSS_PREFIX}
```
sarahraines, line 169: This const can be defined outside of the table so we don't need to pass it as a prop.

src/components/VirtualizedDataTable/VirtualizedDataTable.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  const table = useReactTable<TData>({
+    data: tableData,
+    columns: columnDefs,
+    getCoreRowModel: getCoreRowModel(),
+    getSortedRowModel: getSortedRowModel(),
+  })
+
+  const { rows } = table.getRowModel()
+  const processedData = rows.map(r => r.original)
```
sarahraines, line 78: This will be an unstable reference and could cause unnecessary re-rendering of the table - let's memoize it!
`const processedData = useMemo(() => rows.map(r => r.original), [rows])`

src/components/VirtualizedDataTable/virtualizedDataTable.scss
```diff
@@ -0,0 +1,49 @@
...
+  position: relative;
+  overflow: auto;
+  padding-right: var(--spacing-sm);
+}
+
+.Layer__UI__VirtualizedTable__state-container {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
```
sarahraines, line 11: This can be an `HStack`:
`<HStack align='center' justify='center' ...`

src/components/VirtualizedDataTable/VirtualizedDataTable.tsx
```diff
@@ -0,0 +1,222 @@
...
+      transform: `translateY(${virtualRow.start}px)`,
+    }}
+  >
+    {columns.map(column => (
+      <Cell
+        key={column.id}
+        className={classNames(
+          'Layer__table-cell',
+          `${cssPrefix}__cell`,
+          `Layer__UI__Table-Cell__${componentName}--${column.id}`,
```
sarahraines, line 214: I think we can ditch the `Layer__table` prefixed styles everywhere.
```suggestion
        className={classNames(
          `${cssPrefix}__cell`,
          `Layer__UI__Table-Cell__${componentName}--${column.id}`,
```

src/components/VirtualizedDataTable/VirtualizedDataTable.tsx
```diff
@@ -0,0 +1,222 @@
...
+  useReactTable,
+  getCoreRowModel,
+  getSortedRowModel,
+  createColumnHelper,
+} from '@tanstack/react-table'
+import { Loader } from '../Loader/Loader'
+import type { ColumnConfig, Column } from '../DataTable/DataTable'
+import { Table, TableBody, TableHeader, Column as TableColumn, Row, Cell } from '../ui/Table/Table'
+
+const DEFAULT_ROW_HEIGHT = 34
```
sarahraines, line 14: Seems a bit narrow, but not certain since I'm not looking at it. Maybe when we have another use case, we should create some variants of this (e.g., compact vs. normal) so that we can have more/less information density.

sarahraines (approved):
Nice work, thanks for making this!

PR #758, opened 2025-08-21
Daniel/lay 1688 investigate render failure after successful api request

doneel:
1. Fixing very minor row alignment that I noticed that bothered me.
2. Fixing the issue where the table has at the height for at _exactly_ 20 rows instead of at most.
3. Change the default from 20 to 15 rows. 20 requires a pretty big resolution.
4. Fix schema parsing issue that was causing failures for us in testing.
5. Fix terrible row squishing when narrowing the table. Make sure virtualized table cells don't overflow. 
6. Add a tooltip for description rows. With some real world data, we sometimes have descriptions too big for this table. Moving to a model we use elsewhere of ellipsis+conditional tooltip.
7. Enabling the "source" detail view, which needed backwards compatible schema translation.

### DIFF
--- a/src/components/VirtualizedDataTable/VirtualizedDataTable.tsx
+++ b/src/components/VirtualizedDataTable/VirtualizedDataTable.tsx
@@ -41,6 +41,7 @@ export interface VirtualizedDataTableProps<TData extends { id: string }, TColumn
     EmptyState: React.FC
     ErrorState: React.FC
   }
+  shrinkHeightToFitRows: boolean
   // Virtualization-specific props
   height?: number
   rowHeight?: number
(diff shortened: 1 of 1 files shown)


src/components/ProfitAndLossDetailReport/ProfitAndLossDetailReport.tsx
```diff
@@ -176,15 +166,23 @@ export const ProfitAndLossDetailReport = ({
     [PnlDetailColumns.Description]: {
       id: PnlDetailColumns.Description,
       header: stringOverrides?.descriptionColumnHeader || 'Description',
-      cell: row => <Span ellipsis>{row.source?.displayDescription || row.account.accountSubtype.displayName || '-'}</Span>,
+      cell: row => (
+        <Text
+          as='span'
+          withTooltip={TextUseTooltip.whenTruncated}
```
doneel, line 172: Note that I'm unfortunately going back to using the old `<Text />` component here because it supports this tooltip functionality which, after trying it with real world data, is a must-have. 

I tried porting over tooltip functionality into the new text components but there is a non-trivial refactor to be done there, punting for now.

PR #776, opened 2025-09-03
Daniel/fix amount inputs on journal entries

doneel:
I would suggest that everyone try this out to make sure the behavior is intuitive.

### DIFF
--- a/src/components/Input/AmountInput.tsx
+++ b/src/components/Input/AmountInput.tsx
@@ -11,6 +11,21 @@ export interface AmountInputProps extends Omit<CurrencyInputProps, 'onChange'> {
   badge?: ReactNode
 }
 
+const transformCurrencyValue = (rawValue: string): string => {
+  // Remove non-digit characters except decimal point
+  const cleaned = rawValue.replace(/[^\d.]/g, '')
+
+  // If there are more than 2 digits after decimal, shift them left
+  const parts = cleaned.split('.')
+  if (parts.length === 2 && parts[1].length > 2) {
+    const integerPart = parts[0] + parts[1].slice(0, -2)
+    const decimalPart = parts[1].slice(-2)
+    return `${integerPart}.${decimalPart}`
+  }
+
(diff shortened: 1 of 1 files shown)


src/components/Input/AmountInput.tsx
```diff
@@ -41,6 +41,21 @@ export const AmountInput = ({
       decimalScale={2}
       decimalsLimit={2}
       disableAbbreviations
+      allowDecimals={true}
```
sarahraines, line 44:
```suggestion
      allowDecimals
```

src/components/Input/AmountInput.tsx
```diff
@@ -41,6 +41,21 @@ export const AmountInput = ({
...
+
+        // If there are more than 2 digits after decimal, shift them left
+        const parts = cleaned.split('.')
+        if (parts.length === 2 && parts[1].length > 2) {
+          const integerPart = parts[0] + parts[1].slice(0, -2)
+          const decimalPart = parts[1].slice(-2)
+          return `${integerPart}.${decimalPart}`
+        }
+
+        return cleaned
```
sarahraines, line 57: If you type a second decimal, this only uses the first two parts of the string, e.g., if you had $450.23, and add a decimal in as $4.50.23, the number that comes out is $4.50. I think this is fine, but just calling it out.

src/components/Input/AmountInput.tsx
```diff
@@ -41,6 +41,21 @@ export const AmountInput = ({
       decimalScale={2}
       decimalsLimit={2}
       disableAbbreviations
+      allowDecimals={true}
+      transformRawValue={(rawValue) => {
```
sarahraines, line 45: This function can be defined outside of the component since it's agnostic of any state or props.

sarahraines (approved):
Looks and feels good! This also (inadvertently?) fixed a bug around how the number gets formatting if the user types multiple dollar signs, so yay!

PR #784, opened 2025-09-05
fix: various testing feedback

sarahraines:
## Description
This fixes a number of bugs from demo testing, including:
1. Using our invoice totals utils to calculate sales taxes and discounts on request
2. Table sizing for the invoices table
3. Truncation (with ellipsis) of long invoice numbers across various UIs
4. Table column sizing and truncation (with ellipsis) for long cells


AndrewAcomb (approved):
LGBT

PR #786, opened 2025-09-05
show pending transactions

AndrewAcomb:
## Description
We currently do not show pending transactions (that is, transactions we have not yet assigned suggestions or auto-categorized). 
However, there is no reason to prevent users from categorizing bank transactions without suggestions. It will also make the experience less confusing if/when there are delays generating suggestions.

For now, I decided against doing any special styling on the bank transaction rows themselves. Instead, I'm adding a section in the category dropdown that informs the user that suggestions are currently being generated for the transaction and displays a loading spinner (the same one we use for the tag selector).

Screenshot 2025-09-05 at 1 20 18 PM


src/components/CategorySelect/CategorySelect.tsx
```diff
@@ -148,6 +150,20 @@ const Option = (
     renderInAppLink?: (details: LinkingMetadata) => ReactNode
   },
 ) => {
+  if (props.data.type === 'Loading suggestions') {
```
doneel, line 153: Can we turn this into a type so we don't have to rely on direct string matching?

doneel (approved):

src/components/CategorySelect/CategorySelect.tsx
sarahraines: I was thinking the same thing - maybe we need a new value for `OptionActionType`

src/components/CategorySelect/CategorySelect.tsx
```diff
@@ -308,6 +324,24 @@ export const CategorySelect = ({
...
+          payload: {
+            id: 'loading_suggestions',
+            option_type: OptionActionType.HIDDEN,
+            display_name: 'LOADING SUGGESTIONS',
+            subCategories: null,
+          },
+        } satisfies CategoryOption,
+      ],
+    },
+  ]
```
sarahraines, line 343: Let's use the same "SUGGESTIONS" header as before, and then optionally show either 1) the suggestions if we have them, or 2) the generating suggestions pseudo item.

src/components/CategorySelect/CategorySelect.tsx
```diff
@@ -148,6 +150,20 @@ const Option = (
     renderInAppLink?: (details: LinkingMetadata) => ReactNode
   },
 ) => {
+  if (props.data.type === 'Loading suggestions') {
+    return (
+      <components.Option
```
sarahraines, line 155: Let's make sure this option doesn't darken on hover
Screenshot 2025-09-05 at 1 54 02 PM

PR #787, opened 2025-09-09
Added ability to view account numbers in chart of accounts view and journal view. Added ability to edit account numbers

layer-shyam:
## Description
This PR is to add the ability to the account numbers of ledger accounts in both the chart of accounts view and journal view. You can also edit the account numbers through the edit button in the chart of accounts view

## How this has been tested?
Created a bunch of accounts, added and deleted account numbers from them.


src/components/ChartOfAccountsForm/ChartOfAccountsForm.tsx
```diff
@@ -121,7 +122,8 @@ export const ChartOfAccountsForm = ({
         <div className='Layer__chart-of-accounts__form-edit-entry'>
           <Text weight={TextWeight.bold}>{entry.name}</Text>
           <Text weight={TextWeight.bold}>
-            ${centsToDollars(entry.balance || 0)}
+            $
```
layer-shyam, line 125: Don't ask me why elint did this

PR #1138, opened 2026-01-22
fix: 100ms debounce on responsive component

darrensapalo:
## Description

1. Go to Tax Estimates, set to list view size
2. Use button to toggle between estimates and payments
3. On the estimates page, notice that there is flickering on the component


darrensapalo:
No longer occurs- we've sorted out the flickering.

PR #1290, opened 2026-04-21
Update AGENTS.md styling instructions

jimwei1:
## How this has been tested?

---

> [!NOTE]
> **Low Risk**
> Documentation-only change that updates styling conventions; no runtime behavior or build impact.
> 
> **Overview**
> Updates `AGENTS.md` to formalize CSS guidance by adding a dedicated **CSS Styles** section (discouraging `style` props/utility class strings, defining SCSS file naming, spacing scale, and BEM naming) and a **CSS Selectors** subsection (limits nesting depth, restricts descendant selectors, and forbids targeting other components’ internal classes). Removes the older, shorter style rules block in favor of the expanded guidance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eece12f6aea091eed098082fc67fe0991897ac15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,18 +15,25 @@
   - **`useCallback`**: Only for functions passed as props to `memo()`-wrapped children or used in Hook dependency arrays. Not needed for handlers on plain HTML elements.
   - **`useMemo`**: Only when (a) the computation is expensive (>1ms), (b) the result is an object/array prop to a `memo()`-wrapped child, or (c) the result is used in a Hook dependency array. Never memoize primitives.
 
+# CSS Styles
+
+- Do not use the `style` prop or write raw utility/atomic class strings in `className`
+- Do not re-state styles the target component or its parent stack already set
+- Use design-system props exposed by components instead of a new declaration. For any styling that component props can't express, add a BEM class in PascalCase: `Layer__ComponentName__Element--modifier`
+- Put component rules in a side-effect imported `.scss` file with camelCased component name: `componentName.scss`
+- Use the following spacing scale: (`4xs|3xs|2xs|xs|sm|md|lg|xl|2xl|3xl|5xl`)
+
+## CSS Selectors
+
+- Nest selectors that share a block prefix under one `&__` root; do not nest more than one level (excluding pseudo-classes, pseudo-elements, and media queries): `.Layer__<ComponentName> { &__Element { … } }`
+- Only use descendant combinator selectors when the element can't own a class.
+- Do not target another component's internal classes. Instead, add a prop or exposed class to that component
+
 # Components
 
 - Use `Span` from `@ui/Typography/Text` instead of `<span>`
 - Use `<HStack>` and `<VStack>` from `@ui/Stack/Stack` instead of `<div>`
 - Build on existing components in `@ui` before creating new ones
-
-# Style
-
-- Do not use inline styles
-- Create `.scss` file matching component name but with lowercase first letter and import directly
-- Use design system spacing scale
-- For css class names, use BEM in PascalCase
 
 # Abstractions
 


AGENTS.md
```diff
@@ -15,19 +15,26 @@
   - **`useCallback`**: Only for functions passed as props to `memo()`-wrapped children or used in Hook dependency arrays. Not needed for handlers on plain HTML elements.
   - **`useMemo`**: Only when (a) the computation is expensive (>1ms), (b) the result is an object/array prop to a `memo()`-wrapped child, or (c) the result is used in a Hook dependency array. Never memoize primitives.
 
+# CSS Styles
+
+- Do not use the `style` prop or write raw utility/atomic class strings in `className`
```
darrensapalo, line 20: Agree about not using `style` - though I haven't seen AI agents generate `style` code for me before. No problem adding this in though- it's just one line.

> write raw utility/atomic class strings in `className`

I personally don't know what this means- how does utility or atomic classes differ from the usual classes?

AGENTS.md
```diff
@@ -15,19 +15,26 @@
...
+- Do not re-state styles the target component or its parent stack already set
+- Use design-system props exposed by components instead of a new declaration. For any styling that component props can't express, add a BEM class in PascalCase: `Layer__ComponentName__Element--modifier`
```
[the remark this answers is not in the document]
darrensapalo:
```suggestion
- DRY: use existing design-system props instead defining new SCSS classes. For any styling that component props can't express, follow BEM style: `Layer__Block__Element--modifier` where Block and Element are in PascalCase, and Layer__ is simply a namespace.
```

AGENTS.md
```diff
@@ -15,19 +15,26 @@
...
+- Put component rules in a side-effect imported `.scss` file with camelCased component name: `componentName.scss`
```
darrensapalo, line 23: what's the intention when you say 'side-effect'?

This sounds more clear:
```suggestion
- `ComponentName.tsx` must have `import './componentName.scss'`. Each tsx should have its own scss.
- The scss stylesheets must be in camelCase. Components must be in PascalCase.
```

AGENTS.md
```diff
@@ -15,19 +15,26 @@
...
+- Use the following spacing scale: (`4xs|3xs|2xs|xs|sm|md|lg|xl|2xl|3xl|5xl`)
+
+## CSS Selectors
+
+- Nest selectors that share a block prefix under one `&__` root; do not nest more than one level (excluding pseudo-classes, pseudo-elements, and media queries): `.Layer__<ComponentName> { &__Element { … } }`
+- Only use descendant combinator selectors when the element can't own a class.
+- Do not target another component's internal classes. Instead, add a prop or exposed class to that component
```
darrensapalo, line 30:
```suggestion
```

When building either embeddable components (i.e. components our customers use) and internal reusable components (i.e. components our internal engineers use),  I actually don't advise delegating prop/interface/component design to the AI particularly. In my experience, I feel like the LLM lacks the context, effective tradeoffs, and alignment on how we foresee we might be using components.

For example, some of our guiding principles have been:
- 1️⃣  YAGNI - If we don't need it right now, let's not add a prop.

If there's only one use-case, then it might not need to be a prop. Maybe it's better to be a constant style set for a component rather than something that can be changed.

> - Do not target another component's internal classes. Instead, add a prop or exposed class to that component

The above description I think comes from the context of: Component A (Parent, e.g. a View) is currently being worked on to change styling. It contains Component B (child, a composable element). For some reason (e.g. it doesn't have its own styling file, or whatever), styling Component B is changed within Component A's stylesheet. Thus, Component A "targets another component's (Component B) internal classes".

If the above is the situation you're encountering, I think the solution isn't to just "add a prop or expose class to that component" but instead to think about our code conventions (e.g. each component has its own stylesheet) and intention (use-cases: is this a single use-case? if so constant styling on Component B rather than a prop. If not, explore use-case and determine if a prop is suitable).

Another guiding principle I've observed that we practice in building components is:
- 2️⃣ Be opinionated and product-oriented.

We want to think carefully about the product, how our users use our components and how our engineers compose our elements. So we want to make good default props so that component users (our engineers, our AI agents) already have good recommendations and we build better code quality brick by brick.

Opening this discussion primarily because I care about alignment across our team, and so that alignment bears fruit as clear instructions for the LLMs to execute (clear PASS/FAIL conditions for the LLM's performance).

darrensapalo (commented):
I have some thoughts here Jim!

PR #1327, opened 2026-04-27
Add Canadian Tax Codes v2

jimwei1:
## How this has been tested?

---

> [!NOTE]
> **Medium Risk**
> Adds tax-code handling to categorization/matching payloads and replaces the transaction category store with a new categorization store, touching multiple transaction review surfaces and request builders.
> 
> **Overview**
> **Adds Canadian tax code support to bank transaction categorization.** The UI now surfaces a *Tax Code* selector (including a new mobile drawer) in bulk categorize, single-row categorize, and split flows, and includes `taxCode` in the categorize/split payloads while forcing `null` for *Exclusion* categories.
> 
> **Refactors categorization state to carry more than a category.** The old `BankTransactionsCategoryStoreProvider` is replaced with `BankTransactionsCategorizationStoreProvider`, storing `{ category, taxCode }` per transaction; related hooks, default selection upsert, and bulk match-or-categorize logic are updated accordingly.
> 
> Includes related cleanup: makes `tax_code`/split `taxCode` non-optional in types, moves many bank-transaction helpers to `@utils/bankTransactions/shared`, and updates `ActionableList`/modal styling + CSS naming to align with the design-system conventions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ee303cfa0356ff3d358cd99c320a36df365abab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/utils/bankTransactions/taxCode.ts
```diff
@@ -0,0 +1,56 @@
+import { type BankTransaction, type BankTransactionTaxOption } from '@internal-types/bankTransactions'
+import type { Classification } from '@schemas/categorization'
+
+export type BankTransactionTaxCodeOption = {
+  label: string
+  value: string
+}
```
sarahraines: We should do something like the `ServiceAsOption` class where we can pass the tax code to create one and get back a label/value from it.

src/components/ActionableList/actionableList.scss
```diff
@@ -16,39 +16,35 @@
...
-        color: var(--color-base-500);
-      }
-    }
-
-    &.Layer__actionable-list-item--secondary {
+    &--secondary {
       color: var(--color-base-500);
 
-      &.Layer__actionable-list-item--as-link:hover {
+      &.Layer__actionable-list__item--as-link:hover {
```
darrensapalo: Code convention we should follow for BEM:
- `Layer__ActionableList__Item--asLink`
- `Namespace__BlockNameAsPascalName__ElementNameAsPascalName--modifiersWithCamelCase`
- Block and Element should be `PascalCase` while Modifier should be `camelCase`.

Feel free to confirm with @sarahraines

src/components/ActionableList/actionableList.scss
```diff
@@ -16,39 +16,35 @@
...
   }
 
-  .Layer__actionable-list__select {
+  &__content {
+    max-width: 36ch;
+  }
+
+  &__content-description {
+    color: var(--color-base-500);
+  }
```
darrensapalo: BEM

src/components/ActionableList/ActionableList.tsx
```diff
@@ -32,18 +33,19 @@ export const ActionableList = <T,>({
 }: ActionableListProps<T>) => {
   return (
     <ul className={classNames('Layer__actionable-list', className)}>
-      {options.map((x, idx) => (
+      {options.map(x => (
         <li
           role='button'
           onClick={() => onClick(x)}
-          key={`actionable-list-item-${idx}`}
+          key={x.id}
```
darrensapalo, line 41: I like that we use the object's `id` instead of relying on `idx` because this fixes the assumption that the sorting of the input data is stable. Thanks for fixing. 👍

src/components/BankTransactionRow/bankTransactionRow.scss
```diff
@@ -1,12 +1,14 @@
...
+  }
 
-.Layer__bank-transaction-row__category {
-  overflow: hidden;
-  width: 100%;
+  &__category {
+    overflow: hidden;
+    width: 100%;
+  }
 }
```
darrensapalo, line 14: This is good DRYing of code. I like it.

I know BankTransaction is a huge component- I don't think we should do any major BEM refactoring here- especially because our customers style this component as well.

But for net new components & existing smaller components that are less likely to be styled by our customers, let's make sure our SCSS class names now follow the BEM format.

src/components/BankTransactionRow/BankTransactionRow.tsx
```diff
@@ -82,8 +85,8 @@ export const BankTransactionRow = ({
   const isTransactionSelected = isSelected(bankTransaction.id)
   const { count: bulkSelectionCount } = useCountSelectedIds()
   const isBulkSelectionActive = bulkSelectionCount > 0
-  const { setTransactionCategory } = useBankTransactionsCategoryActions()
-  const { selectedCategory } = useGetBankTransactionCategory(bankTransaction.id)
+  const { setTransactionCategorization } = useBankTransactionsCategorizationActions()
+  const { selectedCategorization } = useGetBankTransactionCategorization(bankTransaction.id)
```
darrensapalo, line 89: Just curious- what's the rationale for the change of pulling out the parent (categorization) instead of the current behavior which is `selectedCategory`? I feel like I'm missing context as to why this was changed. Probably a bug that you're fixing that I can learn about.

src/components/BankTransactions/BankTransactionsListItemCategory/bankTransactionsListItemCategory.scss
```diff
@@ -1,11 +1,10 @@
-.Layer__bankTransactionsListItemCategory__List {
+.Layer__bankTransactionsListItemCategory__List,
+.Layer__bankTransactionsListItemCategory__Mobile {
```
darrensapalo, line 2: I think `__Mobile` is not necessarily a block/element and is used as a modifier. Is this small enough that this can be fixed to follow the codebase's BEM convention? Will leave it up to you and Sarah to decide.

src/components/BankTransactionsMobileList/BankTransactionsMobileListBusinessForm.tsx
```diff
@@ -161,8 +188,19 @@ export const BankTransactionsMobileListBusinessForm = ({
...
+              ariaLabel={t('bankTransactions:label.category', 'Category')}
             />
           )}
+        {showCategorization && showTaxCodeSelector && (
+          <TaxCodeSelectDrawerWithTrigger
+            options={taxCodeOptions}
+            value={selectedCategorization?.taxCode ?? null}
+            onChange={handleTaxCodeChange}
+            isDisabled={selectedCategory?.classification?.type === 'Exclusion'}
+            hasSelection={selectedCategorization?.taxCode !== undefined}
```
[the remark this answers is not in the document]
darrensapalo: I think this may be valid. Should verify.

src/components/BankTransactionsMobileList/bankTransactionsMobileListSplitForm.scss
```diff
@@ -1,8 +1,18 @@
-.Layer__BankTransactionsMobileSplitForm__SplitRow {
-  max-inline-size: 100%;
-}
+.Layer__BankTransactionsMobileSplitForm {
+  &__SplitGridContainer {
+    display: grid;
+    grid-template-columns: 5.5rem 1fr;
+    gap: var(--spacing-xs);
+    align-items: center;
```
darrensapalo, line 6: Can we use HStack or VStack instead of defining these scss? for this:

```
<div className='Layer__BankTransactionsMobileSplitForm__SplitGridContainer'>
```

src/components/BankTransactionsMobileList/bankTransactionsMobileListSplitForm.scss
```diff
@@ -1,8 +1,18 @@
...
+  }
+
+  &__AmountInput {
+    inline-size: 100%;
+    text-align: right;
```
darrensapalo, line 11: I think we can use our Span component which has `align: right` - and we only have 3 usages of this <AmountInput> component (2 of which is new in this PR)- Let's fix that by using our Span instead of adding new scss.

src/components/BankTransactionsMobileList/bankTransactionsMobileListSplitForm.scss
```diff
@@ -1,8 +1,18 @@
...
+  }
 
-.Layer__BankTransactionsMobileSplitForm__AmountInput {
-  flex-shrink: 0;
-  inline-size: 6rem;
+  &__TotalAmountInput {
+    width: 5.5rem;
+    text-align: right;
```
darrensapalo, line 16: Samething here- `text-align: right` available in our `Span`. I think the width is fine here.

src/components/BankTransactionsMobileList/bankTransactionsMobileListSplitForm.scss
```diff
@@ -1,8 +1,18 @@
...
+.Layer__BankTransactionsMobileSplitForm {
+  &__SplitGridContainer {
+    display: grid;
+    grid-template-columns: 5.5rem 1fr;
+    gap: var(--spacing-xs);
+    align-items: center;
+  }
+
+  &__AmountInput {
+    inline-size: 100%;
```
darrensapalo, line 10: Can you confirm if the `inline-size` necessary here?

src/components/BankTransactionsMobileList/BankTransactionsMobileListSplitForm.tsx
```diff
@@ -98,68 +101,156 @@ export const BankTransactionsMobileListSplitForm = ({
...
         && (
           <VStack gap='sm'>
-            <Text weight={TextWeight.bold} size={TextSize.sm}>
-              {t('bankTransactions:action.split_transaction', 'Split transaction')}
-            </Text>
-            <VStack gap='sm'>
-              {localSplits.map((split, index) => (
-                <HStack
-                  key={`split-${index}`}
+            {localSplits.map((split, index) => (
```
darrensapalo: I think Splits are complex enough to be refactored to component of its own, even if it is file-scoped (i.e. not exported)

because the code below has the cases for:
- `greater than 1 split`
- `exactly 1 split`
- `split form has an error`

I think splitting this up (_pun intended_) would help with component architecture & readability. This would then just be:

```ts
{showCategorization ?? (<Splits ... />)
<BankTransactionFormFields ... />
```

src/components/BankTransactionsMobileList/BankTransactionsMobileListSplitForm.tsx
```diff
@@ -98,68 +101,156 @@ export const BankTransactionsMobileListSplitForm = ({
...
+                        <TaxCodeSelectDrawerWithTrigger
+                          options={taxCodeOptions}
+                          value={getSelectedTaxCodeOption(split.taxCode ?? null)}
+                          onChange={handleTaxCodeChange(index)}
+                          isDisabled={
+                            !showCategorization
+                            || split.category?.classification?.type === 'Exclusion'
+                          }
+                        />
+                      </>
```
darrensapalo: I would consider having a file-scoped component for this just to separate the prop calculations (value, isDisabled) from the DOM structure of the tax code selector. This improves readability for a reviewer- separating state calculations vs. DOM structure.

Also- is the `HStack` with the empty `Span` at the start necessary?

src/components/BusinessForm/BusinessFormMobileItem.tsx
```diff
@@ -4,24 +4,26 @@ import ChevronRight from '@icons/ChevronRight'
...
 import { Span } from '@ui/Typography/Text'
-import { type BankTransactionCategoryComboBoxOption } from '@components/BankTransactionCategoryComboBox/bankTransactionCategoryComboBoxOption'
 
 import './businessFormMobileItem.scss'
 
-export type BusinessFormOptionValue = BankTransactionCategoryComboBoxOption
+export interface BusinessFormOptionValue {
+  label: string
+  value: string
+}
```
darrensapalo, line 13: It is a good idea to revisit existing types that have this exact structure, so that we don't make a new type.

As of this PR, there are 13 types that use label and value; Can you please DRY this so that we don't add the 13th?

Image

src/components/CategorySelect/CategorySelectDrawerWithTrigger.tsx
```diff
@@ -22,15 +20,20 @@ export const CategorySelectDrawerWithTrigger = ({ value, onChange, showTooltips
   const [isDrawerOpen, setIsDrawerOpen] = useState(false)
 
   return (
-    <HStack fluid className='Layer__CategorySelectDrawerWithTrigger'>
+    <HStack fluid>
```
darrensapalo, line 23: Great deletion. 💯

darrensapalo (commented):
Next batch of feedback!

src/components/ExpandedBankTransactionRow/utils.ts
```diff
@@ -146,6 +153,11 @@ export const getLocalSplitStateForExpandedTransaction = (
     {
       amount: bankTransaction.amount,
       category: coercedSelectedCategory ?? null,
+      taxCode: coercedSelectedCategory?.classification?.type === 'Exclusion'
+        ? null
+        : selectedCategorization === undefined
+          ? bankTransaction.tax_code ?? null
+          : selectedTaxCode === null ? null : selectedTaxCode.value,
```
darrensapalo: There seems to be repetitive code in terms of determining the taxCode of a `Classification` object or `Split` object across the codebase. 

Search: `classification?.type === 'Exclusion'` and find cases where taxCode is being set.

Is there a way we can have the logic all in the same place, but taking in different objects (Splits, Classification, ephemeral,  intermediate values composed during option selection)? 

This will help with (a) logical consistency, (b) less code so better maintainability, (c) improved readability when these objects are constructed on the fly.

src/components/TaxCodeSelect/TaxCodeSelectDrawer.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { VStack } from '@ui/Stack/Stack'
+import { ActionableList, type ActionableListOption } from '@components/ActionableList/ActionableList'
+import { SearchField } from '@components/SearchField/SearchField'
+
+import { NO_TAX_CODE } from './constants'
+
+export type TaxCodeSelectOption = {
+  label: string
+  value: string
+}
```
darrensapalo: Duplicate of `BankTransactionTaxCodeOption`.

src/components/TaxCodeSelect/TaxCodeSelectDrawer.tsx
```diff
@@ -0,0 +1,103 @@
...
+              if (item.id === NO_TAX_CODE) {
+                onSelect(null)
+              }
+              else {
+                onSelect(item.value)
+              }
+              close()
+            }}
+            selectedId={selectedId}
+            className='Layer__bank-transaction-mobile-list-item__categories_list'
```
darrensapalo, line 99: 1. I am not sure how many elements usually we would expect for this `ActionableList`, but as that number grows, this anonymous function `onClick` will cause the component to repeatedly re-render.

If count is less than 3 or 5, maybe its okay to leave it since the list is a "small component" and rerender cost is low. Otherwise, it makes sense to `useCallback` here.

src/hooks/features/bankTransactions/useUpsertBankTransactionsDefaultCategories.tsx
```diff
@@ -1,19 +1,23 @@
...
-    }))
-
-    setOnlyNewTransactionCategories(defaultCategories)
-  }, [bankTransactions, setOnlyNewTransactionCategories])
+    setOnlyNewTransactionCategorizations(new Map(
+      bankTransactions.map(transaction => [
+        transaction.id,
+        {
+          category: getDefaultSelectedCategoryForBankTransaction(transaction),
+          taxCode: null,
```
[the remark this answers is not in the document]
darrensapalo: ^ I think this is a valid consideration.

src/components/BankTransactionRow/BankTransactionRow.tsx
jimwei1: This is a general refactor to change the store from a map of `<String, Category>` to `<String, Categorization>`, where `Categorization = { Category, TaxCode}`. I think it's worth considering separating into two functions to retrieve / set category and tax code separately, but I wanted to consolidate to make the call sites explicitly set / not set these

src/providers/BankTransactionsCategorizationStore/BankTransactionsCategorizationStoreProvider.tsx
```diff
@@ -0,0 +1,178 @@
+import { createContext, type PropsWithChildren, useContext, useState } from 'react'
+import { createStore, useStore } from 'zustand'
+
+import type { BankTransactionCategoryComboBoxOption } from '@components/BankTransactionCategoryComboBox/bankTransactionCategoryComboBoxOption'
+
+export type TaxCodeSelectOption = {
+  label: string
+  value: string
+}
```
darrensapalo: Duplicate of `BankTransactionTaxCodeOption` and `TaxCodeSelectOption` from `TaxCodeSelectDrawer`.

src/components/BankTransactions/BankTransactionsListItemCategory/bankTransactionsListItemCategory.scss
jimwei1: hm... unsure if clients are modifying this

src/components/BankTransactionsMobileList/bankTransactionsMobileListSplitForm.scss
jimwei1: No, but we could potentially create a new component for a grid layout. This is consistent with how grids are defined elsewhere in repo atm

src/components/BankTransactionsMobileList/bankTransactionsMobileListSplitForm.scss
jimwei1: Do you mean AmountInput? I'm hesitant to get rid of AmountInput but open to adding text-align:right as a prop there

src/components/BankTransactionsMobileList/bankTransactionsMobileListSplitForm.scss
jimwei1: yes

src/components/BusinessForm/BusinessFormMobileItem.tsx
jimwei1: good point, will refactor for them to use `ComboBoxOption` as available

src/components/ExpandedBankTransactionRow/ExpandedBankTransactionRow.tsx
```diff
@@ -286,66 +325,104 @@ export const ExpandedBankTransactionRow = ({
...
+                            </VStack>
+                          )
+                        })}
                       </div>
                       {splitFormError && <HStack pb='sm'><ErrorText>{splitFormError}</ErrorText></HStack>}
-                      <div className={`${className}__total-and-btns`}>
+                      <div
+                        className={classNames(
+                          `${className}__total-and-btns`,
+                          asListItem && `${className}__total-and-btns--list-item`,
```
darrensapalo: Let's avoid classname concatenation here.

Not asking for a rename of `.Layer__expanded-bank-transaction-row__total-and-btns` to BEM format, but at least for the modifier we are adding `--list-item`, let's:

1. add the full class name so it's 1 less thing to dereference/recalculate later when we do the BEM cleanup.
2. Use the BEM format for the modifier at least, even if block and element aren't updated.


```suggestion
                          asListItem && `.Layer__expanded-bank-transaction-row__total-and-btns--listItem`,
```

src/styles/bank_transactions.scss
```diff
@@ -419,6 +415,10 @@
     grid-area: button;
   }
 
+  & > &__tax-code__Container {
```
darrensapalo: Is there a place we're using this right now? I can't seem to find it.

src/styles/bank_transactions.scss
```diff
@@ -878,6 +878,15 @@
...
   }
+
+  .Layer__expanded-bank-transaction-row__total-and-btns--list-item {
+    flex-direction: column;
+    align-items: stretch;
+
+    .Layer__input-tooltip {
+      max-width: 100%;
+    }
+  }
```
darrensapalo: Since this is a new component altogether, I recommend using `Stack` instead of `<div>`

darrensapalo (commented):
Next batch of feedback

src/styles/bank_transactions.scss
jimwei1: kicking the can down the road

src/styles/bank_transactions.scss
jimwei1: yup this is being used - discussed & moved

jimwei1:
Superseded by the stacked draft PR chain:\n\n- #1329\n- #1330\n- #1331\n\nClosing this original broad PR in favor of the split stack.

jimwei1:
Superseded by the stacked replacement PRs:

- #1349
- #1350
- #1351

PR #1404, opened 2026-05-12
fix: isolate task mutation state per task to prevent concurrent submission bugs

doneel:
## Summary

- **Root cause (primary):** All four task mutation hooks (`useSubmitUserResponseForTask`, `useUploadDocumentsForTask`, `useDeleteUploadsOnTask`, `useUpdateTaskUploadDescription`) built their SWR key from `{ businessId, apiUrl, accessToken }` — without `taskId`. Every `TasksListItem` on the page therefore shared the same SWR mutation state slot. When two tasks were submitted in quick succession, SWR's concurrent mutation handling on a shared key could fire both HTTP requests while producing unpredictable client-side state — the rare "two tasks submitted at the same time" bug.
- **Fix:** `taskId` is now a parameter to each hook and is included in `buildKey`, giving every task item its own isolated mutation state.
- **`isMutating` guards:** With per-task keys in place, `isMutating` now accurately reflects only that task's in-flight state. Submit buttons are disabled for the duration of a request, preventing rapid re-clicks.
- **Mobile index keys:** `TasksListMobile` used `key={index}` on both task lists. After submitting task A (which re-sorts to "completed"), React reused the component instance at index 0 for task B while it still held task A's local `userResponse` state. Fixed by using `key={task.id}`.

## Files changed

- `src/hooks/api/businesses/[business-id]/tasks/[task-id]/user-response/useSubmitResponseForTask.ts`
- `src/hooks/api/businesses/[business-id]/tasks/[task-id]/upload/useUploadDocumentsForTask.ts`
- `src/hooks/api/businesses/[business-id]/tasks/[task-id]/upload/delete/useDeleteUploadsOnTask.ts`
- `src/hooks/api/businesses/[business-id]/tasks/[task-id]/upload/update-description/useUpdateTaskUploadDescription.ts`
- `src/components/Tasks/TasksListItem.tsx`
- `src/components/Tasks/TasksListMobile.tsx`

## Test plan

- [ ] Submit a free-response task — confirm single submission, panel closes, task moves to completed
- [ ] Submit a document-upload task — confirm single submission, files shown, task moves to completed
- [ ] Submit two different tasks back-to-back quickly — confirm only one server call per task, no duplicate submissions
- [ ] Verify submit button is disabled while a request is in-flight (can't double-click
- [ ] On mobile: submit the first task in the list, confirm the second task's textarea is empty (not populated with the first task's response)
EOF
)

### DIFF
--- a/src/components/Tasks/TasksListItem.tsx
+++ b/src/components/Tasks/TasksListItem.tsx
@@ -29,10 +29,10 @@ export const TasksListItem = forwardRef<HTMLDivElement, TasksListItemProps>((
   const [userResponse, setUserResponse] = useState(task.user_response ?? '')
   const [selectedFiles, setSelectedFiles] = useState<File[]>()
 
-  const { trigger: handleSubmitUserResponseForTask } = useSubmitUserResponseForTask()
-  const { trigger: handleUploadDocumentsForTask } = useUploadDocumentsForTask()
-  const { trigger: handleDeleteUploadsOnTask } = useDeleteUploadsOnTask()
-  const { trigger: handleUpdateTaskUploadDescription } = useUpdateTaskUploadDescription()
+  const { trigger: handleSubmitUserResponseForTask, isMutating: isSubmittingResponse } = useSubmitUserResponseForTask(task.id)
+  const { trigger: handleUploadDocumentsForTask, isMutating: isUploadingDocuments } = useUploadDocumentsForTask(task.id)
+  const { trigger: handleDeleteUploadsOnTask } = useDeleteUploadsOnTask(task.id)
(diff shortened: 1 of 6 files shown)


doneel:
Superseded by two focused PRs:

- #1447 — the narrow fix (mobile `key={task.id}` + in-flight submit guard)
- #1448 — regression tests + Vitest harness, stacked on #1447

While building a reproduction I confirmed that the `taskId`-in-SWR-mutation-key change from this PR was not actually the fix (SWR's `useSWRMutation` keeps `isMutating`/`data` per hook instance, so a shared key doesn't leak state across tasks). It was dropped for simplicity. The real fix is the React keying plus the `isMutating` submit guard.

PR #1407, opened 2026-05-12
Align Tax Estimate AmountWithLabel Symbols to Amount

jimwei1:
### Before

image

### After

image

---

> [!NOTE]
> **Low Risk**
> Low risk styling/layout-only change to the tax summary equation; no business logic or data handling is modified.
> 
> **Overview**
> Adjusts the `TaxSummaryCard` equation layout so each amount+label stack is center-aligned and given a consistent minimum inline width, improving alignment of totals with the `-`/`=` operators.
> 
> Removes the dedicated operator CSS class from the equation symbols and relies on the base `Span` styling instead.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5b9ed33081a912780a298bb11f245999c51d378. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/components/TaxDetails/TaxSummaryCard/taxSummaryCard.scss
+++ b/src/components/TaxDetails/TaxSummaryCard/taxSummaryCard.scss
@@ -41,7 +41,7 @@
   }
 
   &__AmountWithLabel {
-    min-width: 0;
+    width: 5.5rem;
   }
 
   &__Operator {
--- a/src/components/TaxDetails/TaxSummaryCard/taxSummaryCard.scss
+++ b/src/components/TaxDetails/TaxSummaryCard/taxSummaryCard.scss
@@ -44,10 +44,6 @@
     width: 5.5rem;
   }
 
-  &__Operator {
-    align-self: center;
-  }
-
   &--mobile {
     width: 100%;
   }
--- a/src/components/TaxDetails/TaxSummaryCard/taxSummaryCard.scss
+++ b/src/components/TaxDetails/TaxSummaryCard/taxSummaryCard.scss
@@ -41,7 +41,7 @@
   }
 
   &__AmountWithLabel {
-    width: 5.5rem;
+    min-inline-size: 5.5rem;
   }
 
   &--mobile {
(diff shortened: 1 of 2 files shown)


jimwei1:
@cursor review

src/components/TaxDetails/TaxSummaryCard/taxSummaryCard.scss
```diff
@@ -41,7 +41,7 @@
   }
 
   &__AmountWithLabel {
-    min-width: 0;
+    width: 5.5rem;
```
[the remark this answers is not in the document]
jimwei1: not a problem since it's a desktop component

src/components/TaxDetails/TaxSummaryCard/taxSummaryCard.scss
```diff
@@ -41,11 +41,7 @@
   }
 
   &__AmountWithLabel {
-    min-width: 0;
-  }
-
-  &__Operator {
-    align-self: center;
+    width: 5.5rem;
```
sarahraines: What happens if the amount gets really large here? Is it truncated? I think we need some ability to flex.

jimwei1: There should be enough space well into 7-8 digits. This is close to the max width each `AmountLabel` can get without the label text wrapping - goal is to always leave the max amount of width for each `AmountLabel` to keep the spacing between consistent

image

jimwei1: Although it does start to look weird with certain 9 digit amounts 

image

image

darrensapalo: https://www.loom.com/share/eaad108040204ddf873b2ad094d26b7a

I think changing this to `min-inline-size: 5.5rem;` instead of a fixed width is acceptable.

```suggestion
    min-inline-size: 5.5rem;
```

darrensapalo (approved):
Please use `min-inline-size` but other than that lgtm!

PR #1433, opened 2026-05-25
Create + Update Categorization Rules

AndrewAcomb:
Screenshot 2026-05-25 at 3 49 44 PM

Screenshot 2026-05-25 at 3 50 08 PM

---

> [!NOTE]
> **Medium Risk**
> New upsert and counterparty list APIs drive automatic categorization behavior; creates trigger broad cache reloads (transactions, P&amp;L) that could affect downstream categorization and reporting until data settles.
> 
> **Overview**
> Adds **create and edit** flows for bank transaction categorization rules from the Categorization Rules screen, wired through a drawer form and new API hooks.
> 
> Users get a **Create Rule** action in the header and **edit** on each row (desktop table and mobile list). Opening either launches `CategorizationRuleFormDrawer` with fields for **counterparty** (server-backed search via new `useListCounterparties`), **category**, **direction** (any / money in / out), and optional **min/max amount** filters. Desktop uses comboboxes; mobile uses drawer-based pickers. Submit calls **`useUpsertCategorizationRule`** (POST create / PATCH update) with Effect schema encoding; creates reload rules, bank transactions, and P&amp;L caches, while updates patch the rules list in place.
> 
> The rules list now shows an **Amount** column/row detail via `getCategorizationRuleAmountLabel`. Schemas gain **`PatchCategorizationRuleSchema`**, amount min/max on `CategorizationRule`, relaxed optional fields on create, and **`BankTransactionCounterparty`** typing. **`ComboBox`** accepts `filterOption={null}` for API-driven counterparty search. Shared **`convertCentsToNonRecursiveBigDecimal` / `convertNonRecursiveBigDecimalToCents`** helpers are adopted in invoice and service forms.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b4fdef23d1a018eed59fa9ad23a82464eec9429. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/hooks/api/businesses/[business-id]/categorization-rules/useUpsertCategorizationRule.ts
```diff
@@ -0,0 +1,135 @@
...
+  const stableProxiedTrigger = useCallback(
+    async (
+      arg: UpsertCategorizationRuleArg,
+      options?: Parameters<typeof originalTrigger>[1],
+    ) => {
+      const triggerResult = await originalTrigger(arg, options)
+
+      void forceReloadCategorizationRules()
+
+      if (arg.mode === 'create') {
```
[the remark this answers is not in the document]
Comment (AndrewAcomb) [reply src/hooks/api/businesses/[business-id]/categorization-rules/useUpsertCategorizationRule.ts:114]: we do not retroactively apply rules upon updating, so not needed.

src/components/CategorizationRules/CategorizationRuleForm/CategorizationRuleForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+            return (
+              <HStack className='Layer__CategorizationRuleForm__FormError'>
+                <DataState
+                  icon={<AlertTriangle size={16} />}
+                  status={DataStateStatus.failed}
+                  title={validationErrors[0] || submitError}
+                  titleSize={TextSize.md}
+                  inline
+                />
+              </HStack>
```
darrensapalo: Can you check `src/components/ChartOfAccountsForm/ChartOfAccountsForm.tsx` and `src/components/CustomAccountForm/CustomAccountForm.tsx` for patterns on how individual error messages are rendered on the field, leveraging React Aria's `<FieldError>` component?

I would also check with Sarah if this is the pattern we want to continue using (i.e. follow pattern only) or if there's any extensions/improvements we want to do with regard to form and error validation.

src/components/CategorizationRules/CategorizationRuleForm/CounterpartyComboBox.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [debouncedQuery, setDebouncedQuery] = useState('')
+  const debouncedSetQuery = useDebounce(setDebouncedQuery)
+  const { data, isLoading } = useListCounterparties({
+    q: debouncedQuery || undefined,
+    limit: 50,
+  })
+
+  const onInputValueChange = useCallback((nextInput: string) => {
+    debouncedSetQuery(nextInput)
+  }, [debouncedSetQuery])
```
darrensapalo: When finding using useCallbacks that directly call a function, it's a good idea to verify if the callback being wrapped is unstable (which the useCallback then stabilizes).

In this case, `useDebounce` provides a memoized (stable) debounceSetQuery so useCallback has no benefit. We can directly use `debouncedSetQuery` instead.

src/components/CategorizationRules/CategorizationRuleForm/CounterpartyComboBox.tsx
```diff
@@ -0,0 +1,93 @@
...
+      {showLabel && (
+        <Label size='sm' htmlFor={inputId}>
+          {label}
+        </Label>
+      )}
+      <ComboBox
+        options={options}
+        selectedValue={selectedOption}
+        onSelectedValueChange={option => onValueChange(option?.counterparty ?? null)}
+        onInputValueChange={onInputValueChange}
```
darrensapalo:
```suggestion
        onInputValueChange={debouncedSetQuery}
```

src/components/CategorizationRules/CategorizationRuleForm/CounterpartyComboBox.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+type CounterpartyComboBoxProps = {
+  label: string
+  value: BankTransactionCounterparty | null
+  onValueChange: (counterparty: BankTransactionCounterparty | null) => void
+  showLabel?: boolean
+  isReadOnly?: boolean
+  isError?: boolean
+  placeholder?: string
+  className?: string
```
darrensapalo: If this is only used in one place, I would recommend we don't expose `className` as a configurable property, and only expose it as a prop when we find the need to configure it differently in another place.


```suggestion
```

After looking at the code, it seems this prop is actually also not used- so we might as well remove it if unused.

src/components/CategorizationRules/CategorizationRuleForm/CounterpartyComboBox.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const selectedOption = useMemo(() => {
+    if (!value) return null
+    return options.find(option => option.value === value.id) ?? null
+  }, [options, value])
+
+  const additionalAriaProps = !showLabel ? { 'aria-label': label } : {}
+
+  return (
+    <VStack gap='3xs' className={className}>
```
darrensapalo:
```suggestion
    <VStack gap='3xs'>
```

src/components/CategorizationRules/CategorizationRuleForm/formUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+  category: Classification | null
+  bankDirectionFilter: DirectionFormValue
+  amountMinFilter: NonRecursiveBigDecimal | null
+  amountMaxFilter: NonRecursiveBigDecimal | null
+}
+
+const centsToFormAmount = (cents: number | null | undefined): NonRecursiveBigDecimal | null => {
+  if (cents == null) return null
+  return toNonRecursiveBigDecimal(convertCentsToBigDecimal(cents))
+}
```
darrensapalo, line 36: We seem to be doing this operation in other places on the codebase (see [this](https://github.com/Layer-Fi/layer-react/blob/d56f11628e4ec83e7db5987276a7976ce910eeb5/src/components/Invoices/InvoiceForm/formUtils.ts#L49) on _another_ form util as well).

I wonder if `centsToFormAmount` could be generalized not just for the form amount, but consolidated as a single helper for all operations from 'cents' to 'non recursive big decimal' and reused across the app not naming tied to form amounts.

These two frequently go together:

```
toNonRecursiveBigDecimal(convertCentsToBigDecimal( --- ))
```

src/components/CategorizationRules/CategorizationRuleForm/CategorizationRuleForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+    { value: BankDirectionFilter.MONEY_IN, label: t('common:label.money_in', 'Money in') },
+    { value: BankDirectionFilter.MONEY_OUT, label: t('common:label.money_out', 'Money out') },
+  ]
+
+  return (
+    <Form className='Layer__CategorizationRuleForm' onSubmit={blockNativeOnSubmit}>
+      <form.Subscribe selector={state => state.errorMap}>
+        {(errorMap) => {
+          const validationErrors = flattenValidationErrors(errorMap)
+          if (validationErrors.length > 0 || submitError) {
```
darrensapalo, line 45: Image

Can we think of a better empty state when the user has 0 counterparties, rather than allowing them to arrive on a form that they think they can fill up, when they technically cannot do any operation because they are still in an empty state?

Not sure if the better experience here is disabling the button to open the form and providing a tooltip explaining why they can't yet- or if it's an empty state on the slide in component. There's mobile UX considerations here as well.

src/hooks/api/businesses/[business-id]/counterparties/useListCounterparties.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  usePreserveInfiniteSize(swrResponse)
+
+  return new ListCounterpartiesSWRResponse(swrResponse)
+}
+
+const withUpdatedCounterparty = (updated: BankTransactionCounterparty) =>
+  (rule: BankTransactionCounterparty): BankTransactionCounterparty => rule.id === updated.id ? updated : rule
+
+export function useCounterpartiesGlobalCacheActions() {
```
Comment (darrensapalo) [inline src/hooks/api/businesses/[business-id]/counterparties/useListCounterparties.ts]: This seems like an unused hook. Do we expect any use-cases where we'll need it? If not, let's remove unused code.

darrensapalo (approved):
Some revisions and some questions, but overall looks great!

src/components/CategorizationRules/CategorizationRuleForm/CategorizationRuleForm.tsx
AndrewAcomb: TimeEntryForm and InvoiceForm both use this same banner pattern, so I'd lean toward keeping it for consistency.  

Happy to switch if we decide as a team to standardize on inline <FieldError> — probably worth a follow-up PR either way.

src/components/CategorizationRules/CategorizationRuleForm/CategorizationRuleForm.tsx
AndrewAcomb: Screenshot 2026-05-26 at 4 34 48 PM

Updated - added specific copy for this edge case

src/components/CategorizationRules/CategorizationRuleForm/formUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+  values: CategorizationRuleFormValues,
+  t: TFunction,
+) => {
+  const errors: Array<Record<string, string>> = []
+
+  if (!values.counterparty) {
+    errors.push({
+      counterparty: t('categorizationRules:validation.counterparty_required', 'Counterparty is required.'),
+    })
+  }
```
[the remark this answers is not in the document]
AndrewAcomb: Accepting this - we do not intend to support description-only rules long-term

src/components/CategorizationRules/CategorizationRuleForm/formUtils.ts
AndrewAcomb: Created `convertCentsToNonRecursiveBigDecimal`

src/components/CategorizationRules/CategorizationRuleForm/CategorizationRuleForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+export const CategorizationRuleForm = ({ formState, onSuccess }: CategorizationRuleFormProps) => {
+  const { t } = useTranslation()
+  const { form, submitError } = useCategorizationRuleForm({ formState, onSuccess })
+
+  const blockNativeOnSubmit = useCallback((e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault()
+    e.stopPropagation()
+  }, [])
+
+  const directionOptions: Array<{ value: DirectionFormValue, label: string }> = [
```
sarahraines: Let's memoize this since it's a non-primitive value being passed as a prop.

src/components/CategorizationRules/CategorizationRuleForm/CategorizationRuleForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+            showLabel
+          />
+        )}
+      </form.Field>
+
+      <form.AppField name='bankDirectionFilter'>
+        {field => (
+          <field.FormRadioGroupField
+            label={t('common:label.direction', 'Direction')}
+            orientation='horizontal'
```
sarahraines: I'd slightly prefer the vertical orientation here.
```suggestion
            orientation='vertical'
```

src/components/CategorizationRules/CategorizationRuleForm/CategorizationRuleFormDrawer.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { VStack } from '@ui/Stack/Stack'
+import { CategorizationRuleForm } from '@components/CategorizationRules/CategorizationRuleForm/CategorizationRuleForm'
+import { type CategorizationRuleFormState } from '@components/CategorizationRules/CategorizationRuleForm/formUtils'
+
+export type CategorizationRuleFormDrawerProps = {
+  isOpen: boolean
+  onOpenChange: (isOpen: boolean) => void
+  onSuccess: (rule: CategorizationRule) => void
+  formState: CategorizationRuleFormState | null
+}
```
sarahraines, line 17: Some marginally more elegant TypeScript that uses a discriminated union to indicate that if the drawer is open, formState must be defined.

```suggestion
type CategorizationRuleFormDrawerSharedProps = {
  onOpenChange: (isOpen: boolean) => void
  onSuccess: (rule: CategorizationRule) => void
}

export type CategorizationRuleFormDrawerProps = CategorizationRuleFormDrawerSharedProps &
  (
    | { isOpen: false; formState: null }
    | { isOpen: true; formState: CategorizationRuleFormState }
  )
```

src/components/CategorizationRules/CategorizationRuleForm/formUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+    name: null,
+    category: values.category,
+    suggestion1: null,
+    suggestion2: null,
+    suggestion3: null,
+    businessNameFilter: null,
+    clientNameFilter: null,
+    merchantTypeFilter: null,
+    transactionDescriptionFilter: null,
+    transactionTypeFilter: null,
```
sarahraines: Do we need to pass all of these null values to the API, or can we just omit them?

src/components/CategorizationRules/utils.ts
```diff
@@ -23,3 +23,28 @@ export const getCategorizationRuleDirectionLabel = (
...
+    })
+  }
+  if (min != null) {
+    return t('categorizationRules:label.amount_at_least', '≥ {{amount}}', {
+      amount: formatCurrencyFromCents(min),
+    })
+  }
+  return t('categorizationRules:label.amount_at_most', '≤ {{amount}}', {
+    amount: formatCurrencyFromCents(max as number),
+  })
```
sarahraines: These probably don't need to be wrapped in translation blocks, since there are no words.

src/hooks/api/businesses/[business-id]/categorization-rules/useUpsertCategorizationRule.ts
```diff
@@ -0,0 +1,135 @@
...
+  const { trigger: originalTrigger } = mutationResponse
+
+  const stableProxiedTrigger = useCallback(
+    async (
+      arg: UpsertCategorizationRuleArg,
+      options?: Parameters<typeof originalTrigger>[1],
+    ) => {
+      const triggerResult = await originalTrigger(arg, options)
+
+      void forceReloadCategorizationRules()
```
Comment (sarahraines) [inline src/hooks/api/businesses/[business-id]/categorization-rules/useUpsertCategorizationRule.ts]: I believe we should patch the cache if we're updating rather than force reloading the whole list, unless updating a rule affects other rules.

src/hooks/api/businesses/[business-id]/categorization-rules/useUpsertCategorizationRule.ts
```diff
@@ -0,0 +1,135 @@
...
+        accessToken,
+        {
+          params: { businessId, categorizationRuleId: arg.categorizationRuleId },
+          body: arg.body,
+        },
+      ).then(decode)
+    },
+    {
+      revalidate: false,
+    },
```
Comment (sarahraines) [inline src/hooks/api/businesses/[business-id]/categorization-rules/useUpsertCategorizationRule.ts:102]:
```suggestion
      revalidate: false,
      throwOnError: true,
    },
```

src/components/CategorizationRules/CategorizationRuleForm/CounterpartyComboBox.tsx
```diff
@@ -0,0 +1,108 @@
...
+  onValueChange,
+  showLabel,
+  isReadOnly,
+  isError,
+  placeholder,
+}: CounterpartyComboBoxProps) => {
+  const { t } = useTranslation()
+  const inputId = useId()
+  const [debouncedQuery, setDebouncedQuery] = useState('')
+  const debouncedSetQuery = useDebounce(setDebouncedQuery)
```
sarahraines: You can use `useDebouncedSearchInput` here.

src/components/CategorizationRules/CategorizationRuleForm/CounterpartyComboBox.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const fetchedOptions = useMemo<ReadonlyArray<CounterpartyOption>>(() => {
+    if (!data) return []
+    return data.flatMap(({ data: page }) => page.map(counterpartyToOption))
+  }, [data])
+
+  const slots = useMemo(() => {
+    if (debouncedQuery === '' && !isLoading && fetchedOptions.length === 0) {
+      return {
+        EmptyMessage: (
```
sarahraines: I think we should always have some `EmptyMessage` - it will only show if there are truly no options, and we can case the language on whether or not we're filtering (e.g., no results vs no filtered results).

We should also include an `ErrorMessage`.

src/components/CategorizationRules/CategorizationRuleForm/CounterpartyComboBox.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { useListCounterparties } from '@hooks/api/businesses/[business-id]/counterparties/useListCounterparties'
+import { useDebounce } from '@hooks/utils/debouncing/useDebounce'
+import { ComboBox } from '@ui/ComboBox/ComboBox'
+import type { ComboBoxOption } from '@ui/ComboBox/types'
+import { VStack } from '@ui/Stack/Stack'
+import { Label, Span } from '@ui/Typography/Text'
+
+type CounterpartyOption = ComboBoxOption & {
+  counterparty: BankTransactionCounterparty
+}
```
sarahraines: We should create CategoryComboBoxOption class to handle the logic of converting the counterparty to a combobox option. See `ReportComboBoxOption` for an example.

src/components/CategorizationRules/CategorizationRuleForm/CategorizationRuleFormDrawer.tsx
```diff
@@ -0,0 +1,61 @@
...
+  isOpen,
+  onOpenChange,
+  onSuccess,
+  formState,
+}: CategorizationRuleFormDrawerProps) => {
+  const { t } = useTranslation()
+
+  const title = formState?.mode === 'edit'
+    ? t('categorizationRules:action.edit_rule', 'Edit rule')
+    : t('categorizationRules:action.new_rule', 'New rule')
```
sarahraines:
```suggestion
    ? t('categorizationRules:action.edit_rule', 'Edit Rule')
    : t('categorizationRules:action.new_rule', 'New Rule')
```

src/components/CategorizationRules/CategorizationRulesView/ResponsiveCategorizationRulesView.tsx
```diff
@@ -53,32 +54,53 @@ const CategorizationRulesErrorState = () => {
...
+        <Heading size='sm'>{t('categorizationRules:label.categorization_rules', 'Categorization Rules')}</Heading>
+      </HStack>
+      <HStack pie='md' align='center' gap='xs'>
+        <Button variant='outlined-light' onPress={onCreateRule}>
+          {t('categorizationRules:action.new_rule', 'New rule')}
+          <Plus size={16} />
         </Button>
-      )}
-      <Heading size='sm'>{t('categorizationRules:label.categorization_rules', 'Categorization Rules')}</Heading>
+      </HStack>
```
sarahraines, line 78: Primary button, title case, more instructive content.
```suggestion
      <HStack pie='md' align='center' gap='xs'>
        <Button onPress={onCreateRule}>
          {t('categorizationRules:action.create_rule', 'Create Rule')}
          <Plus size={16} />
</Button>
      </HStack>
```

src/components/CategorizationRules/CategorizationRuleForm/CategorizationRuleFormDrawer.tsx
```diff
@@ -0,0 +1,61 @@
...
+      isOpen={isOpen}
+      onOpenChange={onOpenChange}
+      aria-label={title}
+      slots={{ Header }}
+    >
+      <VStack pb='lg'>
+        <VStack pi='md'>
+          <CategorizationRuleForm formState={formState} onSuccess={onSuccess} />
+        </VStack>
+      </VStack>
```
sarahraines: There's a bit of weird extra vertical space between the header of the drawer and the first form element.

src/components/CategorizationRules/CategorizationRuleForm/CategorizationRuleFormDrawer.tsx
```diff
@@ -0,0 +1,61 @@
...
+    : t('categorizationRules:action.new_rule', 'New rule')
+
+  const Header = useCallback(({ close }: { close: () => void }) => (
+    <CategorizationRuleFormDrawerHeader title={title} close={close} />
+  ), [title])
+
+  if (formState === null) return null
+
+  return (
+    <Drawer
```
sarahraines, line 56: Do we have plans for mobile compatibility here? E.g., using the mobile drawer?

src/components/CategorizationRules/CategorizationRuleForm/CategoryMobileDrawer.tsx
```diff
@@ -0,0 +1,90 @@
...
+type CategoryMobileDrawerProps = {
+  label: string
+  value: Classification | null
+  onValueChange: (value: Classification | null) => void
+  showLabel?: boolean
+  isReadOnly?: boolean
+  placeholder?: string
+}
+
+export const CategoryMobileDrawer = ({
```
sarahraines, line 19: I think this may be duplicative with `CategorySelectDrawerWithTrigger`, which we can add extra props (such as label) to if necessary.

src/components/CategorizationRules/CategorizationRuleForm/CounterpartyMobileDrawer.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const options = useMemo<ReadonlyArray<CounterpartyComboBoxOption>>(() => {
+    if (!value) return fetchedOptions
+    if (fetchedOptions.some(option => option.value === value.id)) return fetchedOptions
+    return [new CounterpartyComboBoxOption(value), ...fetchedOptions]
+  }, [fetchedOptions, value])
+
+  const selectedOption = useMemo(() => {
+    if (!value) return null
+    return options.find(option => option.value === value.id) ?? null
+  }, [options, value])
```
sarahraines: This logic seems to be the same between the mobile drawer and the combobox - can we extract it into a hook?

src/components/CategorizationRules/CategorizationRuleForm/CounterpartyComboBox.tsx
```diff
@@ -0,0 +1,117 @@
...
+        <Span>{message}</Span>
+      </VStack>
+    )
+
+    let EmptyMessage
+    if (isListError) {
+      EmptyMessage = wrapMessage(
+        t('categorizationRules:error.load_counterparties', 'Couldn’t load counterparties. Please try again.'),
+      )
+    }
```
sarahraines: Easier/more conventional to do something like the following, then wrap it at the end.
```suggestion
    let emptyMessageContent = ''
    if (isListError) {
      emptyMessageContent = t('categorizationRules:error.load_counterparties', 'Couldn’t load counterparties. Please try again.')
    }
```

src/components/CategorizationRules/CategorizationRuleForm/CounterpartyComboBox.tsx
```diff
@@ -0,0 +1,117 @@
...
+    else {
+      EmptyMessage = wrapMessage(
+        t('categorizationRules:empty.no_matching_counterparties', 'No matching counterparties.'),
+      )
+    }
+
+    return {
+      EmptyMessage,
```
sarahraines: Related to above comment.
```suggestion
      EmptyMessage: (
        <VStack pi='md'>
          <Span>{emptyMessageContent}</Span>
        </VStack>
      ),
```

src/components/CategorizationRules/CategorizationRulesView/ResponsiveCategorizationRulesView.tsx
```diff
@@ -181,6 +207,14 @@ export const ResponsiveCategorizationRulesView = () => {
...
         useDrawer={isMobile}
       />
+      {formState && (
+        <CategorizationRuleFormDrawer
+          isOpen
+          formState={formState}
+          onOpenChange={onFormDrawerOpenChange}
+          onSuccess={onFormSuccess}
+        />
+      )}
```
sarahraines: I think we prefer passing the `isOpen` prop to ensure that the animation works correctly when the drawer comes in and out. We may need to reset the form state to default values though between times that we open the drawer (e.g., for different rules, or same rule but reopened).
```suggestion
      <CategorizationRuleFormDrawer
        isOpen={!!formState}
        formState={formState}
        onOpenChange={onFormDrawerOpenChange}
        onSuccess={onFormSuccess}
      />
```

sarahraines (approved):

PR #1557, opened 2026-07-06
tests: clean up createMutationHook (rename args, drop isEnabled) + test swr hook factories

sarahraines:
## Scope
Cleanup of the SWR hook factories, focused on `createMutationHook`:

1. **Clearer argument names in `createMutationHook`** — disambiguates key param names from resolved values (`keyParamValues` / `KeyParamValues`; generic `TKeyParamNames`).

2. **Removed `isEnabled` from mutation options.** It gates a *query's* automatic fetch, but a mutation only runs on `trigger()` — so it never gated anything; it just nulled the key, making a stray `trigger()` throw SWR's cryptic `"missing key"`. Dropped it from the factory and the ~10 callers that passed it (they keep passing the id, defaulting to `''` where the value was previously guarded).
   - Follow-up fix: guarded `UnifiedReportDownloadButton` (the one caller that used `isEnabled` as a real "don't run without a report" guard) so it fails fast and is disabled while there's no active report.

3. **Fixed `tags` leaking into mutation request params.** The mutation fetcher only stripped `accessToken`/`apiUrl`, so the internal `tags` key field reached `request`. Now stripped too, matching `createKeyedFetcher` (the query path).

4. **Tests for the three factories** — `createQueryHook`, `createMutationHook`, `createInfiniteQueryHook` (39 tests) covering every config option and edge case: auth/businessId forwarding + key-context stripping, `isEnabled` (queries/infinite), schema decode / passthrough / decode-failure, `select`, `swrOptions` precedence, `isLocalized`, mutation body/param derivation, and infinite pagination (cursor threading, `flattenedData`, `hasMore`, `keyDefaults`, terminal `fetchMore` no-op).

## Test infra added
- `LayerTestProvider` gains a `locale` prop (+ exported `TEST_LAYER_ACCESS_TOKEN`).
- `renderHookWithAuth` — renders in `LayerTestProvider` and awaits auth before returning, so assertions never race a missing key (no timers).
- `getRequestOptions(request, i)` — typed shorthand for `request.mock.calls[i][2]`.

## Notes
- `node_modules/react` was not installed locally (react is a dev+peer dep); `npm install` was required to run the suite.

## Verification
`npx vitest run --dir src/hooks/utils/swr` → 39 passing; `tsc --noEmit` and eslint clean.

---

> [!NOTE]
> **Medium Risk**
> Shared mutation factory behavior changes across many API hooks; the unified report download path is the main user-visible guard change, mitigated by explicit UI checks and new factory tests.
> 
> **Overview**
> Cleans up **`createMutationHook`**: renames key-param generics/args for clarity, **removes `isEnabled`** (it only nulled the mutation key and did not gate `trigger()`), and **stops `tags` from being forwarded** into HTTP `params` (aligned with the query fetcher).
> 
> **Call-site updates:** upsert/delete hooks no longer pass `isEnabled`; update paths still pass ids (often `''` in create mode). **`useDeleteTimeEntry`** now requires a definite `timeEntryId`. **`useUnifiedReportExcel`** drops `isEnabled`; **`UnifiedReportDownloadButton`** early-returns and disables the button when there is no active report so export is not triggered against an empty route.
> 
> Adds **39 tests** for `createQueryHook`, `createMutationHook`, and `createInfiniteQueryHook`, plus test helpers (`renderHookWithAuth`, `getRequestOptions`, `LayerTestProvider` locale + `TEST_LAYER_ACCESS_TOKEN`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b8fbb239d766ec9dd243a80014b7e7a554c6fc36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/components/UnifiedReports/UnifiedReportDownloadButton.tsx
+++ b/src/components/UnifiedReports/UnifiedReportDownloadButton.tsx
@@ -27,13 +27,14 @@ export function UnifiedReportDownloadButton({ icon }: UnifiedReportDownloadButto
   const emitLayerEvent = useEmitLayerEvent(LayerEventComponent.UnifiedReports)
 
   const onPress = () => {
-    if (report) {
-      emitLayerEvent({
-        type: LayerEventType.ReportsDownloadClicked,
-        version: 1,
-        payload: { reportKey: report.key },
-      })
-    }
+    // No active report means no export route/params — don't fire a request against an empty route.
+    if (!report) return
+
+    emitLayerEvent({
+      type: LayerEventType.ReportsDownloadClicked,
+      version: 1,
+      payload: { reportKey: report.key },
+    })
     void trigger()
   }
 
@@ -47,7 +48,7 @@ export function UnifiedReportDownloadButton({ icon }: UnifiedReportDownloadButto
         variant='outlined'
         onPress={onPress}
         isPending={isMutating}
-        isDisabled={isMutating}
+        isDisabled={isMutating || !report}
         icon={resolvedIcon}
         aria-label={resolvedIcon ? buttonText : undefined}
       >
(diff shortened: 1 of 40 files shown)


tbustama (approved):
Waterzooi

PR #1571, opened 2026-07-09
LAY-2342 Journal Date Filter [5/n] [feat] Support All Time date preset

jeffreylayer:
# Description

Add a context-dependent "All Time" preset whose range spans from the business activation date to the present. 

## Changes

- The date store now stores both the `DateRange` and the `DatePreset` together. When one is set, the other one is derived.
- The date store exposes separate actions for setting and reading the `DateRange` and `DatePreset`
- The store is only constructed when the initial date range can be fully resolved, making sure that the initialization happens with the correct date range (as opposed to instantiating with a sentinel default value)
- The store renders a scoped fallback loading state while the business context loads.

---

> [!NOTE]
> **Medium Risk**
> Changes shared global date filtering and initialization timing; incorrect preset/range derivation could shift report/query windows, though behavior is heavily tested.
> 
> **Overview**
> Adds an **All Time** report period from the business activation date through today, and makes **date preset** a first-class part of the global date store instead of inferring it only in the combo box.
> 
> The Zustand date store now keeps `preset` with `startDate`/`endDate`: `setDatePreset` derives the range (using activation date for All Time), and `setDateRange` / picker changes re-derive the preset (including **Custom** when nothing matches). Store actions thread optional activation date for clamping and All Time resolution. **All Time** initialization waits on business context—the provider renders a **fallback** until the range resolves and errors if All Time is used outside that context; the combo disables All Time while activation is unknown.
> 
> Date selection UI is wired through `useGlobalDatePreset` / `setDatePreset`. `createScopedStore` now takes `createStore` on the **Provider** so each mount can build the store with a fully resolved initial range/preset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1419da07147a3a91a39e116236681b72d03df85f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/components/DateSelection/GlobalDateRangeSelection.tsx
```diff
@@ -1,19 +1,24 @@
...
-  const dateRange = useGlobalDateRange({ dateSelectionMode: 'full' })
+export const GlobalDateRangeSelection = ({ showLabels = false, isCompact = false, includeAllTime = true }: GlobalDateRangeSelectionProps) => {
+  const { startDate, endDate, preset } = useGlobalPresetRange({ dateSelectionMode: 'full' })
   const { setDateRange } = useGlobalDateRangeActions()
+  const { setPresetRange } = useGlobalPresetRangeActions()
 
   return (
     <DateRangeSelection
-      dateRange={dateRange}
+      dateRange={{ startDate, endDate }}
```
sarahraines: This will need to be memoized. We may want to consider this shape instead so it does not need to be re-memoized in the component!

`const { dateRange, preset } = useGlobalPresetRange({ dateSelectionMode: 'full' })`

jeffreylayer: Update the shape to ensure it's stable by splitting it out and not destructing it.

src/providers/DateStoreProvider/internal/types.ts
```diff
@@ -1,11 +1,22 @@
...
-  setDateRange: (options: { startDate: Date, endDate: Date }) => DateRange
-  setMonth: (options: { startDate: Date }) => DateRange
-  setYear: (options: { startDate: Date }) => DateRange
-  setMonthByPeriod: (options: { monthNumber: number, yearNumber: number }) => DateRange
+  /** Set an explicit date range; the preset it represents is derived. */
+  setDateRange: (range: DateRange, activationDate?: Date) => DateRangeWithPreset
+  /** Set an explicit date preset; the concrete date range is derived. */
+  setDatePreset: (datePreset: Exclude<DatePreset, DatePreset.Custom>, activationDate?: Date) => DateRangeWithPreset
+  /** Set an explicit date; the date range and presetit represents is derived. */
+  setDate: (date: Date, activationDate?: Date) => DateRangeWithPreset
```
jeffreylayer: @sarahraines I saw that the `options: { startDate: Date }` argument shape wasn't being leveraged in the current code, so I tentatively updated the function signatures to accept a `date: Date` directly for simplicity and YAGNI purposes.

Let me know what you think! If there are other considerations where using `options` is the right call, I can change it back.

src/utils/date/dateRangePresets.ts
```diff
@@ -112,7 +133,9 @@ export function presetForDateRange(input: DateRange, selectedPreset: DatePreset
     ]),
   )
 
-  if (selectedPreset !== null && selectedPreset !== DatePreset.Custom) {
+  // AllTime is not a relative candidate (its range needs context), so it can't be
+  // derived here — a caller that tracks the stored preset supplies it directly.
+  if (selectedPreset !== null && selectedPreset !== DatePreset.Custom && selectedPreset !== DatePreset.AllTime) {
```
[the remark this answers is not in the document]
jeffreylayer: This is intentional - renaming the function to make this more explicit

src/utils/date/dateRangePresets.ts
```diff
@@ -81,12 +90,24 @@ function fromEntriesStrict<K extends PropertyKey, V>(
...
 }
 
+/**
+ * The `AllTime` range: from the business activation date to the present. Unlike
+ * `rangeForPreset` this needs the activation date, so it is resolved wherever the
+ * business context is available (the store resolver and the preset combo box).
+ */
+export function rangeForAllTime(activationDate: Date): DateRange {
+  return {
+    startDate: startOfDay(activationDate),
```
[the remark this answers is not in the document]
jeffreylayer: This is fixed - I added the correctDateRange helper before returning

src/providers/DateStoreProvider/internal/createScopedDateStore.test.tsx
```diff
@@ -95,9 +106,86 @@ describe('createScopedDateStore', () => {
...
+      { wrapper: Wrapper },
+    )
+
+    // Fallback while the business (and thus the activation date) is still loading.
+    expect(result.current).toBeNull()
+
+    await waitFor(() => expect(result.current).not.toBeNull())
+
+    expect(result.current.datePreset).toBe(DatePreset.AllTime)
+    expect(result.current.range).toEqual(rangeForAllTime(makeBusiness().activationAt))
```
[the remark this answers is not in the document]
jeffreylayer: Fixed

src/components/DateSelection/DateSelectionComboBox.tsx
```diff
@@ -35,25 +32,25 @@ export const DateSelectionComboBox = ({ dateRange, setDateRange, showLabel = fal
       { value: DatePreset.LastQuarter, label: t('date:label.last_quarter', 'Last Quarter') },
       { value: DatePreset.ThisYear, label: t('date:label.this_year', 'This Year') },
       { value: DatePreset.LastYear, label: t('date:label.last_year', 'Last Year') },
+      { value: DatePreset.AllTime, label: t('date:label.all_time', 'All Time') },
```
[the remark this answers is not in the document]
jeffreylayer: Fixed this by only checking the deterministic start time

src/utils/date/dateRangePresets.ts
```diff
@@ -102,23 +130,56 @@ const sameDateRange = (a: DateRange, b: DateRange) =>
...
+ */
+export function derivePresetFromDateRange(
+  input: DateRange,
+  previousPreset: DatePreset | null = null,
+  activationDate?: Date,
+): DatePreset {
+  if (activationDate && sameDateRange(normalize(input, activationDate), normalize(rangeForAllTime(activationDate), activationDate))) {
+    return DatePreset.AllTime
+  }
+  return deriveRelativePresetFromDateRange(input, previousPreset, activationDate) ?? DatePreset.Custom
```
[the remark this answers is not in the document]
jeffreylayer: Fixed

src/providers/DateStoreProvider/internal/buildDateStore.ts
```diff
@@ -1,63 +1,88 @@
...
       return next
     }
 
-    const setDate = ({ date }: { date: Date }): DateRange => {
-      // Always clamp to start of month for date.
-      const monthRange = getDateRange({ mode: 'month', endDate: date })
-      const fullRange = getDateRange({ mode: 'full', startDate: date, endDate: date })
-      return apply({ startDate: monthRange.startDate, endDate: fullRange.endDate })
+    const transformAndSet = (dateRange: DateRange, activationDate?: Date): DateRangeWithPreset => {
+      const correctedRange = correctDateRange(dateRange)
```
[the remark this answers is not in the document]
jeffreylayer: Fixed

src/providers/DateStoreProvider/internal/dateStoreUtils.ts
```diff
@@ -39,16 +39,14 @@ export function getDateRange(options: GetDateRangeOptions): DateRange {
   }
 }
 
-export function withCorrectedRange<TDateRange extends DateRange, TOut>(fn: (options: TDateRange) => TOut) {
-  return (options: TDateRange) => {
-    const { startDate, endDate } = options
+export function correctDateRange(range: DateRange): DateRange {
```
sarahraines: Any reason for the refactor here?

src/providers/DateStoreProvider/internal/types.ts
sarahraines: The reasoning here (across all actions) was to force the callers to use named params,  so it's clear as to what's being passed. This is a bit of a code style personal preference for me, but I think it becomes more relevant as the number of arguments grows.

src/components/DateSelection/DateSelectionComboBox.tsx
```diff
@@ -1,31 +1,36 @@
...
+}: DateSelectionComboBoxProps) => {
   const { t } = useTranslation()
-  const [lastPreset, setLastPreset] = useState<DatePreset | null>(null)
-  const { business } = useLayerContext()
 
-  const selectedPreset = presetForDateRange(dateRange, lastPreset, getActivationDate(business))
+  // AllTime resolves against the business activation date. Until the business
+  // loads (`activationDate == null`), selecting it would be a no-op, so block the
+  // selection until the activationDate is loaded
+  // In practice, the business loads in a fraction of a second, so this is rarely visible.
```
sarahraines: As an overall comment, I think we can cut down on 80% of the comments in the PR, and only include them when it's not self-evident what the code is going.

src/providers/DateStoreProvider/internal/buildDateStore.ts
```diff
@@ -1,63 +1,86 @@
...
+
+      const dateRangeToApply = {
+        startDate: startOfMonth,
+        endDate: endDateClamped,
+      }
 
-    const setMonth = ({ startDate }: { startDate: Date }): DateRange => {
-      return apply(getDateRange({ mode: 'month', endDate: startDate }))
+      return transformAndSet(dateRangeToApply, activationDate)
     }
```
sarahraines: I think this is where `withCorrectedRange` was nice.

src/utils/date/dateRangePresets.ts
```diff
@@ -102,23 +130,66 @@ const sameDateRange = (a: DateRange, b: DateRange) =>
...
+  if (activationDate && sameDateRange(normalize(input, activationDate), normalize(rangeForAllTime(activationDate), activationDate))) {
+    return DatePreset.AllTime
+  }
+  const relativePreset = deriveRelativePresetFromDateRange(input, previousPreset, activationDate)
+
+  if (relativePreset) {
+    return relativePreset
+  }
+
+  if (activationDate && sameDateRange(normalize(input, activationDate), normalize(rangeForAllTime(activationDate), activationDate))) {
```
sarahraines: How does this differ from what's above? Can we also define some of these normalized date ranges as constants so it's easier to figure out what they represent? We may also want to include "All time" in the previous preset calculation, since it can technically have an equivalent range to some of the other presets.

src/providers/DateStoreProvider/internal/buildDateStore.ts
```diff
@@ -1,63 +1,86 @@
...
-  return createStore<DateStore>((set) => {
-    const apply = (next: DateRange): DateRange => {
+export function buildDateStore({ initialRange, initialPreset }: BuildDateStoreOptions) {
+  return createStore<DateStore>((set, get) => {
+    const setDateRangeWithPreset = (range: DateRange, preset: DatePreset): DateRangeWithPreset => {
+      const next = { startDate: range.startDate, endDate: range.endDate, preset }
       set(next)
       return next
     }
+    const transformAndSet = (dateRange: DateRange, activationDate?: Date): DateRangeWithPreset => {
```
sarahraines: Needs more descriptive function name

src/providers/DateStoreProvider/internal/types.ts
jeffreylayer: I agree with that reasoning! Let me change this back to adhere to the existing code-style and for improved readability.

src/providers/DateStoreProvider/internal/buildDateStore.ts
jeffreylayer: I agree that in this specific case, the declarative style of having the higher-order-function style `withCorrectedRange` reads a bit more succinctly at the call-site. 

I still think the improvement in readability and simplicity of the plain function implementation is worth the change. It's only usage is in this call-site, and I think the current function is still relatively readable.

The main advantage for the simple function implementation is less conceptual overhead / typing complexity and more straightforward and readable unit tests.

Happy to talk more about this if you feel strongly about the original implementation!

src/components/DateSelection/DateSelectionComboBox.tsx
jeffreylayer: Sounds good! As someone who's newer to the codebase, I definitely erred on the side of having more generous comments for anything that felt non-obvious to me (or an important behavior decision point).

Reading through some of these comments again, I agree that some of them are probably unnecessary and too verbose. I'll take a pass through and try to remove anything I think should be removed, but please feel free to lmk if I need to take another pass.

src/utils/date/dateRangePresets.ts
jeffreylayer: Great catch! This is definitely an AI-typo accidentally slipping through during a refactor. Let me take another pass at this function and improve its readability and correctness.

src/providers/DateStoreProvider/internal/dateStoreUtils.ts
jeffreylayer: Yes! This was an intentional choice, and the main reason was just to simplify the function's interface for human readability and unit testability.

The helper function does a very simple job, so I wanted to keep its implementation lightweight and easy to unit test. The higher order function implementation felt a bit more syntactically complex without much functional gain. The call-sites using the HOF might be slightly more succinct, but I erred on the side of "simple and plain" rather than fancy.

One of the stronger supporting points for the improved simplicity would be to check out how much simpler the unit tests for this function became :)

Another reason is that the original HOF implementation would've been harder to extend, since the function signature need to match the caller's arguments. With more trailing arguments that needed to be added to `setDateRange`, like `activationDate`, this helper would've needed to add some variadic parameters in its argument types to continue to support the use-case. I felt like the HOF implementation unnecessarily coupled its shape (input params and output) to the caller, whereas a clean pure functional helper felt more decoupled and a better abstraction.

The variadic version would've looked something like this:

```
export function withCorrectedRange<TDateRange extends DateRange, TArgs extends unknown[], TOut>(
  fn: (options: TDateRange, ...rest: TArgs) => TOut,
) {
  return (options: TDateRange, ...rest: TArgs) => {
    const { startDate, endDate } = options

    if (startDate > endDate) {
      return fn({ ...options, startDate: endDate, endDate: startDate }, ...rest)
    }

    return fn(options, ...rest)
  }
}
```

src/providers/DateStoreProvider/internal/buildDateStore.ts
jeffreylayer: Agreed. Let me rename it.

src/providers/DateStoreProvider/internal/types.ts
jeffreylayer: After I reverted this, I got reminded more concretely for why I made this change in the first place. We're fully aligned on principle, but this is a case where I felt that readability improved by un-naming the params. 

For now, I'm totally ok with reverting this to how it was in this PR. This is a stylistic preference that we can chat about later, and we could always make a follow-up PR to implement it if we're on the same page.

I think it'd be better to scope the stylistic named-param decision out into a separate PR anyways, for single-concern / atomic PR purposes.

src/providers/DateStoreProvider/internal/types.ts
```diff
@@ -1,11 +1,17 @@
...
-  setDate: (options: { date: Date }) => DateRange
-  setDateRange: (options: { startDate: Date, endDate: Date }) => DateRange
-  setMonth: (options: { startDate: Date }) => DateRange
-  setYear: (options: { startDate: Date }) => DateRange
-  setMonthByPeriod: (options: { monthNumber: number, yearNumber: number }) => DateRange
+  /** Sets an explicit date range; the preset it represents is derived. */
+  setDateRange: (options: { startDate: Date, endDate: Date }, activationDate?: Date) => DateRangeWithPreset
+  /** Sets an explicit date preset; the concrete date range is derived. */
+  setDatePreset: (options: { datePreset: SelectableDatePreset }, activationDate?: Date) => DateRangeWithPreset
+  setDate: (options: { date: Date }, activationDate?: Date) => DateRangeWithPreset
```
jeffreylayer: @sarahraines 

Would you rather I put the `activationDate` inside the options object?

jeffreylayer: I'm going to assume yes for now and push up a change to implement this. I can revert it if that's not the right direction.

src/providers/DateStoreProvider/internal/buildDateStore.ts
```diff
@@ -1,63 +1,86 @@
...
+      const endDateClamped = getDateRange({ mode: 'full', startDate: date, endDate: date }).endDate
+
+      const dateRangeToApply = {
+        startDate: startOfMonth,
+        endDate: endDateClamped,
+      }
 
-    const setMonth = ({ startDate }: { startDate: Date }): DateRange => {
-      return apply(getDateRange({ mode: 'month', endDate: startDate }))
+      return setDateRangeWithDerivedPreset(dateRangeToApply, activationDate)
```
sarahraines: I think this is a slight readability win
```suggestion
      const monthRange = getDateRange({ mode: 'month', endDate: date })
      const fullRange = getDateRange({ mode: 'full', startDate: date, endDate: date })
      
      const startOfMonth = monthRange.startDate
      const clampedEndDate = fullRange.endDate

      const effectiveDateRange = { startDate: startOfMonth, endDate: clampedEndDate }
      
      return setDateRangeWithDerivedPreset(effectiveDateRange, activationDate)
```

src/providers/DateStoreProvider/internal/buildDateStore.ts
```diff
@@ -1,63 +1,87 @@
...
+    }
+
+    const setMonthByPeriod = (
+      { monthNumber, yearNumber, activationDate }: { monthNumber: number, yearNumber: number, activationDate?: Date },
+    ): DateRangeWithPreset => {
+      const monthIndex = Math.min(Math.max(monthNumber, 1), 12) - 1
+      const firstDayOrMonth = new Date(yearNumber, monthIndex, 1)
+      const dateRangeToApply = getDateRange({ mode: 'month', endDate: firstDayOrMonth })
+      return setDateRangeWithDerivedPreset(dateRangeToApply, activationDate)
+    }
```
sarahraines, line 61: Can we call down to `setMonth` here like the prior implementation? It ensures that the underlying logic remains the same, the only difference being the month period.

src/providers/DateStoreProvider/internal/dateStoreUtils.ts
sarahraines: Ok, I see! I think this is an opportunity for clearer naming then.
```suggestion
export function maybeInvertDateRange(range: DateRange): DateRange {
```

sarahraines (approved):
LGTM! Three comments here.

src/providers/DateStoreProvider/internal/buildDateStore.ts
jeffreylayer: I like it

src/providers/DateStoreProvider/internal/buildDateStore.ts
jeffreylayer: Yes, good catch.

PR #1582, opened 2026-07-10
DRAFT: LAY-2342 Journal Date Filter [6/n] [refactor] Support All Time date preset (Phase 2 — Reorder providers)

jeffreylayer:
# Description

Flip the provider order so BusinessProvider (which provides LayerContext) is the ancestor of GlobalDateStoreProvider. The date store's resolver can now read the business activation date structurally, enabling the AllTime preset even for the global store without special placement.

Because BusinessProvider now sits above the store it can no longer read it to populate LayerContext.dateRange (a public API — useLayerContext is exported). A new LayerContextDateRangeBridge, mounted below the store, re-injects the store-backed dateRange onto the context, so useLayerContext().dateRange keeps working for embedding applications. No circular dependency: business flows down into the resolver, the resolved range flows back onto the context only below the store.

The global default stays ThisMonth (synchronous, non-blocking). Setting the global default to AllTime is now possible but would show an app-wide loading state until the business loads — hence kept opt-in.

## How this has been tested?

---

> [!NOTE]
> **Medium Risk**
> Touches core provider composition and the exported embedding API for global date range; behavior should stay equivalent for consumers under the bridge, but ordering mistakes could break date filters app-wide.
> 
> **Overview**
> **Reorders the React provider tree** so `BusinessProvider` (and `LayerContext`) wraps `GlobalDateStoreProvider` instead of the opposite. `GlobalDateStoreProvider` is removed from `LayerProvider` and nested inside `BusinessProvider`, so the date store can read business (e.g. activation date) from context when resolving presets such as **All Time** for the global store.
> 
> Because `LayerContext` is now above the store, `BusinessProvider` no longer wires `dateRange` from the store directly; it supplies a **This Month placeholder** on the outer context. A new **`LayerContextDateRangeBridge`** sits under the store, merges the store-backed `range` / `setRange` back onto `LayerContext`, and preserves the public **`useLayerContext().dateRange`** API for embedders without a circular dependency.
> 
> Tests cover that the bridge exposes the store range and that `setRange` updates the global store.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 254bcecc2b86c9ca87a4178d4de4827ff73adadf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/providers/BusinessProvider/BusinessProvider.tsx
+++ b/src/providers/BusinessProvider/BusinessProvider.tsx
@@ -5,14 +5,17 @@ import {
   type ColorsPaletteOption,
   type LayerContextAction,
   LayerContextActionName as Action,
+  type LayerContextDateRange,
   type LayerContextValues,
   type LayerThemeConfig,
 } from '@internal-types/layerContext'
 import { errorHandler, type LayerError } from '@utils/api/errorHandler'
 import { buildColorsPalette } from '@utils/colors'
+import { DatePreset, rangeForPreset } from '@utils/date/dateRangePresets'
 import { useAccountingConfiguration } from '@hooks/api/businesses/[business-id]/accounting-config/useAccountingConfiguration'
 import { useBusiness } from '@hooks/api/businesses/[business-id]/useBusiness'
-import { useGlobalDateRange, useGlobalDateRangeActions } from '@providers/DateStoreProvider/GlobalDateStoreProvider'
+import { LayerContextDateRangeBridge } from '@providers/BusinessProvider/LayerContextDateRangeBridge'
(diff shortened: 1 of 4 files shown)


jeffreylayer:
Closing in favor of the non-reordered approach (Alternative 2). Phase 1 (#1571) already mounts the global date store above BusinessProvider and exports `dateRange`/`setRange` via LayerContext, so the provider reorder + LayerContextDateRangeBridge in this PR are unnecessary. The AllTime placement invariant is now enforced by a runtime guard on Phase 1. Recoverable at 254bcecc2 if needed.

PR #1583, opened 2026-07-10
feat: deprecate P&L summaries size variants and gate cashflow footers on breakdown

sarahraines:
## Scope

**`variants` is deprecated and no longer honored.** Both `ProfitAndLossSummariesProps.variants` and `ProfitAndLossSummariesSlotProps.variants` are marked `@deprecated` (pointing consumers at overriding the `--text-*` font size variables instead) and the prop is now ignored, matching how `vertical` is handled. The internal plumbing is fully unwound: `SummariesContent`, `ProfitAndLossSummariesSummary`, `ProfitAndLossSummariesMiniChart`, and `TransactionsToReview` no longer take or thread `variants`, and the mini chart uses its former default dimensions as constants. The Storybook `variants.size` control is removed from the shared summaries controls since it would be a no-op.

**Cash flow footers are gated solely on `showProfitAndLossBreakdown`.** With `{ type: 'cashflow', showProfitAndLossBreakdown: false }`, no tile renders a footer — previously the net tile still rendered the uncategorized-transactions badge when `onTransactionsToReviewClick` was set. Breakdown-on behavior is unchanged.

**Tile labels are spans, not headings.** The labels ("Revenue", "Money in", …) rendered as hardcoded `<h3>`s via `ProfitAndLossSummariesHeading` — but they name a single value rather than a content section, and a fixed heading level inside an embeddable library can't respect host page hierarchies. They now render as `Span size='sm' weight='bold'` (visually identical: 12px bold), and the wrapper component is deleted with the span inlined at its two call sites.

## Verification

- Typecheck, eslint, and all 191 unit tests pass
- Verified in Storybook: cash flow with breakdown off renders no footers and no badge even with the review callback set; breakdown on is unchanged; tile labels render as `<span>` with identical styling (zero heading elements in the summaries subtree)

---

> [!NOTE]
> **Medium Risk**
> Public behavior changes for cash-flow summaries when breakdown is disabled (footers/badges no longer appear on tiles), and integrators passing `variants` will see sizing differences if they relied on `lg`/`sm`.
> 
> **Overview**
> **`variants` on P&L Summaries is deprecated and ignored.** The prop stays on the public types with `@deprecated` guidance (use `--text-*` CSS variables for sizing), but it is no longer threaded through `SummariesContent`, summary tiles, mini charts, or `TransactionsToReview`. Mini charts always use the former default dimensions; README and Storybook drop the `variants.size` control.
> 
> **Cash flow tile footers now depend only on `showProfitAndLossBreakdown`.** When breakdown is off, no tile renders a footer—including the net tile, which previously could still show the transactions-to-review badge via `onTransactionsToReviewClick`. P&L mode still adds the separate **Transactions to review** list item when that callback is set.
> 
> **Tile labels are no longer `<h3>` headings.** `ProfitAndLossSummariesHeading` is removed; labels render as bold `Span` at the same visual size for better embeddability in host page heading hierarchies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b3fd6d6fb9816844004f00a95c402d1fade6a03. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AndrewAcomb (approved):
lgtm!

PR #1615, opened 2026-07-20
Add real-backend Storybook mode

sarahraines:
## Summary

- add a local-only `storybook:real` mode backed by a real Layer business
- isolate real/mock provider selection in a dedicated Storybook provider
- disable MSW and the fixture clock when real mode is active
- document required environment variables and ignore local environment files

## Why

The existing Storybook setup always used fixture data and shifted system time. This makes it possible to inspect the existing stories against real backend data without changing the deterministic mocked workflow used by default.

## Impact

`npm run storybook` remains mock-backed. Developers can copy `.storybook/.env.example` to `.env`, provide a temporary business access token, and run `npm run storybook:real`.

Real-backend credentials must not be used in a published static Storybook build.

## Validation

- `npm run typecheck`
- ESLint for changed Storybook/config files
- `npm run storybook:build -- --quiet`
- real-mode Storybook static build with placeholder credentials

---

> [!NOTE]
> **Medium Risk**
> Adds server-side OAuth token minting and CI deploy wiring with secrets, but sandbox-only token scope and non-prefixed env vars keep credentials out of the client bundle; mock Storybook is unchanged.
> 
> **Overview**
> Adds an opt-in **real-backend Storybook** path (`npm run storybook:real` / `storybook:build:real`) where MSW, fixture clock, and pinned date ranges are turned off and stories run against a live sandbox business via `LayerProvider` and a toolbar **business ID** (with recent-ID history).
> 
> A Vercel **`/api/storybook-token`** handler mints short-lived client-credentials tokens server-side (sandbox/staging only; production refused), proxied from the Storybook dev server in local dev. Stories that make sense without MSW are tagged **`real-backend`** and included when `STORYBOOK_SCOPE=real` for the Vercel static build.
> 
> PR previews are automated: applying the **`storybook-preview`** label triggers a workflow that deploys to Vercel, aliases a stable per-PR hostname, and posts the URL. **`Business`** gains optional **`legal_name`** for the real-backend badge; default mock Storybook behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 203c6f6926db32ca06e493f4464b17b7cb9a6614. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/storybook-preview.yml
```diff
@@ -0,0 +1,129 @@
+name: Storybook Preview
+
+# A label rather than a `/storybook` comment: applying one already requires write access, so GitHub
+# does the authorizing. This holds VERCEL_TOKEN and the repository is public.
```
sarahraines:
```suggestion
```

darrensapalo (approved):

.storybook/.env.example
```diff
@@ -0,0 +1,10 @@
+# Copy to the repository root as `.env`, then run `vercel dev` alongside `npm run storybook:real`.
+
+STORYBOOK_LAYER_BACKEND=real
+STORYBOOK_LAYER_TOKEN_ENDPOINT=/api/storybook-token
```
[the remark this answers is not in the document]
sarahraines: Fixed in [a23c52566](https://github.com/Layer-Fi/layer-react/commit/a23c5256676afe5cdbf58725c38d6f126532e424). Correct, and the worst of the three — Vite always loads the root `.env`, so following the instructions in that very file made `npm run storybook` real too. The line was also redundant: `storybook:real` and `storybook:build:real` already export it. Removed, with a note saying why it must not go back.

src/components/features/linkedAccounts/LinkedAccounts/LinkedAccounts.stories.tsx
```diff
@@ -107,12 +107,12 @@ export default meta
...
 
 export const Default: Story = {
-  tags: ['public-api', 'docs-screenshot'],
+  tags: ['public-api', 'docs-screenshot', 'real-backend'],
 }
 
 // Runs after the meta loader, so it flags the two accounts that survived the trim.
 export const ConfirmingAccounts: Story = {
-  tags: ['public-api'],
+  tags: ['public-api', 'real-backend'],
```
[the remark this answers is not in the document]
sarahraines: Addressed differently in [a23c52566](https://github.com/Layer-Fi/layer-react/commit/a23c5256676afe5cdbf58725c38d6f126532e424) — keeping the stories, making real mode support them.

You're right that the `loaders` are inert with MSW off, and right that my disqualifier rule missed them (it checked story-level `msw:` and `play:` only). But the linked-account flows are the ones most worth exercising live, so rather than dropping them: real mode now keeps the real `react-plaid-link` instead of aliasing in the mock. The mock exists because Plaid's iframe can't run against MSW; against a live backend the iframe works, `sandbox` already implies `usePlaidSandbox`, and the mock's fake public token would be rejected anyway.

The caveat you identified still holds — the loaders do nothing, so those stories show the business's actual state and their names can overclaim. That's now documented in `.storybook/SKILL.md` alongside the same note for `BookkeepingDisabled`.

.github/workflows/storybook-preview.yml
```diff
@@ -0,0 +1,136 @@
...
+        if [ -z "$url" ]; then echo "No deployment URL in output" >&2; exit 1; fi
+
+        # Via the REST API rather than `vercel alias`, which resolves the token to a user and 404s
+        # for this one. `teamId` here is explicit, so there is no user to look up. Best-effort: a
+        # working deployment shouldn't be reported as failed just because aliasing was.
+        alias="layer-react-storybook-pr-$PR.vercel.app"
+        api="https://api.vercel.com"
+        auth="Authorization: Bearer $VERCEL_TOKEN"
+        dpl=$(curl -fsS "$api/v13/deployments/${url#https://}?teamId=$VERCEL_ORG_ID" \
+          -H "$auth" | jq -r '.id // empty')
```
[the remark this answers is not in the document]
sarahraines: Fixed in [a23c52566](https://github.com/Layer-Fi/layer-react/commit/a23c5256676afe5cdbf58725c38d6f126532e424). Correct — and a regression I introduced one commit earlier while fixing the previous alias failure. `run:` uses `bash -e`, and the assignment sat outside the `if`, so a failed lookup aborted the step before `url` reached `GITHUB_OUTPUT`: a good preview reporting failed and never commented, exactly what the best-effort branch was meant to prevent. The assignment is now inside the `if`, which exempts it from `errexit`.

.storybook/SKILL.md
```diff
@@ -115,6 +115,10 @@ state a real backend produces on its own. Two disqualifiers: **story-level
 handlers are fine — they give way to real data), and **a `play` function**, which clicks a row real
 data may not have. `STORYBOOK_SCOPE=real` filters to it.
 
+Story-level `loaders` that seed mock stores are not disqualifying, but they do nothing in real mode —
+those stories show the business's actual state, so the name may overclaim. Real mode also keeps the
+real `react-plaid-link`, so the linked-account flows are live against Plaid sandbox.
+
```
sarahraines: Remove this comment

.storybook/main.ts
```diff
@@ -4,7 +4,9 @@ import { type Alias, type AliasOptions } from 'vite'
 
 import { PUBLIC_API_TAG, REAL_BACKEND_TAG } from './tags'
 
-// Plaid's hosted iframe can't run in Storybook; the mock fakes a successful link.
+// Plaid's hosted iframe can't run against MSW, so the mock fakes a successful link. Real mode keeps
+// the real hook: the iframe works there, `sandbox` implies `usePlaidSandbox`, and the mock's fake
+// public token would be rejected by a live backend anyway.
```
sarahraines: This comment didn't need to increase in size.

.github/workflows/storybook-preview.yml
```diff
@@ -92,10 +92,12 @@ jobs:
         alias="layer-react-storybook-pr-$PR.vercel.app"
         api="https://api.vercel.com"
         auth="Authorization: Bearer $VERCEL_TOKEN"
-        dpl=$(curl -fsS "$api/v13/deployments/${url#https://}?teamId=$VERCEL_ORG_ID" \
-          -H "$auth" | jq -r '.id // empty')
-
-        if [ -n "$dpl" ] && curl -fsS -X POST \
+        # The lookup is assigned inside the `if`, which exempts it from `bash -e`: outside, a failed
+        # lookup would abort the step before `url` reached GITHUB_OUTPUT.
```
sarahraines: No need for this comment

.storybook/SKILL.md
sarahraines: Removed in [be6889720](https://github.com/Layer-Fi/layer-react/commit/be6889720a6c40f8c45926005cd36ebaa6ea591a).

.storybook/main.ts
sarahraines: Back to its original length in [be6889720](https://github.com/Layer-Fi/layer-react/commit/be6889720a6c40f8c45926005cd36ebaa6ea591a) — only the wording changed, from "can't run in Storybook" to "can't run against MSW", since the mock no longer applies in real mode.

.github/workflows/storybook-preview.yml
sarahraines: Removed in [be6889720](https://github.com/Layer-Fi/layer-react/commit/be6889720a6c40f8c45926005cd36ebaa6ea591a).

.storybook/manager.ts
```diff
@@ -0,0 +1,131 @@
...
+            onHide()
+          }),
+        }),
+      },
+      h('button', {
+        'type': 'button',
+        'aria-label': 'Recent businesses',
+        'style': TRIGGER_STYLE,
+      }, '▾'),
+    ),
```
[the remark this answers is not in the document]
sarahraines: False positive — checked against the installed 10.4.6, then in a browser.

`closeOnOutsideClick` is still declared on `WithTooltipPureProps` in `storybook/dist/components/index.d.ts`, with its own doc comment and `@default false`. `trigger` comes through from `react-popper-tooltip`'s `Config`, which that interface extends — the `Omit<…, 'trigger'>` on the same line applies to `TargetContainer`'s DOM prop, not to the config. Both typecheck.

Behaviour, driven headlessly against a real-mode build with seeded history:

- hover alone: does **not** open, so `trigger: 'click'` is in effect (popper defaults to hover)
- click: opens
- click outside: closes, so `closeOnOutsideClick` is in effect

The hover result is the decisive one: were the prop ignored, hovering would have opened it.

PR #1644, opened 2026-07-29
feat: add closing-action ledger entry type and source

doneel:
## Summary
- Add `CLOSING_ACTION_RECOGNIZE_EXTERNAL_REVENUE` to `EntryType` and decode `Closing_Action_Ledger_Entry_Source` in the ledger entry source union
- Render closing-action source details (action type, closing date) in the journal entry side panel, with linking metadata via `EntityName.ClosingAction`
- Add fixture coverage and decode tests for the known closing-action payload

## Test plan
- [ ] Open Chart of Accounts → click a closing-action journal entry line; panel should decode and show entry details + line items (source badge/details present)
- [ ] Confirm Storybook / MSW fixtures include a closing-action ledger entry
- [ ] `npx vitest run src/schemas/generalLedger/ledgerEntry.test.ts`

---

> [!NOTE]
> **Low Risk**
> Additive schema and UI for a new ledger source type; no changes to auth, payments, or existing entry handling beyond extending unions and switch cases.
> 
> **Overview**
> Adds first-class support for **closing-action** journal entries so they decode, link, and render like other ledger source types.
> 
> The API shape is wired through **`CLOSING_ACTION_RECOGNIZE_EXTERNAL_REVENUE`** on `EntryType` and a new **`Closing_Action_Ledger_Entry_Source`** variant on the ledger entry source union (with `closing_action_id`, `action_type`, `closing_date`). In-app navigation uses **`EntityName.ClosingAction`** via `convertLedgerEntrySourceToLinkingMetadata`.
> 
> The journal entry side panel shows **action type** (humanized) and **closing date** for this source. MSW/generated fixtures include a sample closing-action entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89146160879e8a85bcf55ed1ed695431783c1210. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1645, opened 2026-07-29
feat: tolerate unknown ledger entry and agent enum values

doneel:
## Summary
- Add `createOpenEnumSchema` so Effect schemas can accept unknown string values while keeping known enum members for autocomplete
- Wire `entry_type` and `agent` through it so future backend values no longer fail the entire journal entry decode
- Extend decode tests to assert unknown entry types and agents pass through as raw strings

Stacked on #1644.

## Test plan
- [ ] Confirm an entry with a known closing-action `entry_type` still decodes and renders
- [ ] Confirm an unknown `entry_type` / `agent` no longer surfaces as a journal-entry panel error (entry still loads; type humanizes from the raw string)
- [ ] `npx vitest run src/schemas/generalLedger/ledgerEntry.test.ts`
- [ ] After #1644 merges, retarget this PR to `main` if GitHub does not do so automatically

---

> [!NOTE]
> **Low Risk**
> Narrow schema-layer change for forward compatibility; known values still decode the same and only non-strings are rejected.
> 
> **Overview**
> Introduces **`createOpenEnumSchema`** so Effect decoders accept any string for enum-like API fields while keeping known members in the type union for autocomplete. Unknown values pass through as raw strings instead of failing decode.
> 
> **`ClassifierAgent`** and **`EntryType`** on ledger entries now use this helper (replacing strict `Schema.Enums`), so new backend `entry_type` / `agent` values no longer break loading the whole journal entry.
> 
> Adds **`src/schemas/utils.test.ts`** covering open enums, transformed enums with fallbacks, and **`UnwrappedDataResponseSchema`** unwrap/encode behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd6186c67eeec60acd98e780f6cdcce90c136d3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/schemas/utils.ts
+++ b/src/schemas/utils.ts
@@ -13,13 +13,15 @@ export const UnwrappedDataResponseSchema = <A, I, R>(
     },
   )
 
-// Helper function to create enum schemas that accept values the backend may add
-// before this client knows about them, preserving the raw value for display.
-export const createOpenEnumSchema = <T extends Record<string, string>>(enumObject: T) =>
-  Schema.Union(
-    Schema.Enums(enumObject),
-    Schema.String as Schema.Schema<string & NonNullable<unknown>, string>,
-  )
+// `string & {}` accepts any string without collapsing the union, preserving autocomplete
+// for the known enum members.
+type AnyString = string & {}
+type OpenEnum<T extends string> = T | AnyString
+
+// Accepts values the backend may add before this client knows about them, preserving
+// the raw value for display.
+export const createOpenEnumSchema = <T extends Record<string, string>>(_enumObject: T) =>
+  Schema.String as Schema.Schema<OpenEnum<T[keyof T]>>
 
 // Helper function to create transformed enum schemas with safe defaults.
 export const createTransformedEnumSchema = <T extends Record<string, string>>(
--- a/src/schemas/utils.ts
+++ b/src/schemas/utils.ts
@@ -13,6 +13,14 @@ export const UnwrappedDataResponseSchema = <A, I, R>(
     },
   )
 
+// Helper function to create enum schemas that accept values the backend may add
+// before this client knows about them, preserving the raw value for display.
+export const createOpenEnumSchema = <T extends Record<string, string>>(enumObject: T) =>
+  Schema.Union(
+    Schema.Enums(enumObject),
+    Schema.String as Schema.Schema<string & NonNullable<unknown>, string>,
+  )
+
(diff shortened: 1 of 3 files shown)


src/schemas/utils.ts
```diff
@@ -13,6 +13,14 @@ export const UnwrappedDataResponseSchema = <A, I, R>(
     },
   )
 
+// Helper function to create enum schemas that accept values the backend may add
+// before this client knows about them, preserving the raw value for display.
+export const createOpenEnumSchema = <T extends Record<string, string>>(enumObject: T) =>
+  Schema.Union(
+    Schema.Enums(enumObject),
+    Schema.String as Schema.Schema<string & NonNullable<unknown>, string>,
+  )
```
sarahraines: Could we simplify this to a string schema with an explicit open-enum type? 

Since the schema intentionally accepts every string, including `Schema.Enums(enumObject)`, the runtime union doesn’t add any validation—the Schema.String branch already accepts the enum values.

Using `OpenEnum<T>` is a bit clearer - accept future backend values while preserving autocomplete for the known enum values.

Also, a bit of TypeScript, `string & {}` accepts any string while preventing TypeScript from collapsing the entire union of the enums and string to a string value, which will preserve editor autocomplete,

```suggestion
type AnyString = string & {}
type OpenEnum<T extends string> = T | AnyString

export const createOpenEnumSchema = <
  T extends Record<string, string>,
>(_enumObject: T) =>
  Schema.String as Schema.Schema<OpenEnum<T[keyof T]>>
```

src/schemas/generalLedger/ledgerEntry.test.ts
```diff
@@ -69,6 +69,18 @@ describe('LedgerEntrySchema', () => {
...
+
+    expect(entry.entryType).toBe('SOME_FUTURE_ENTRY_TYPE')
+  })
+
+  it('preserves agents this client does not know about', () => {
+    const entry = decodeEntry({ ...CLOSING_ACTION_ENTRY, agent: 'SOME_FUTURE_AGENT' })
+
+    expect(entry.agent).toBe('SOME_FUTURE_AGENT')
+  })
 })
```
sarahraines: Let's test `createOpenEnumSchema` itself in `utils.test.ts` rather than for this specific schema.

sarahraines (approved):

PR #1744, opened 2026-08-11
[LAY-2619] Enabling deep linking

tbustama:
## Description

Adds an optional `initialState` prop to `UnifiedReports` so consumers can deep link to a specific report instead of always landing on P&L. Public API change. Shaped as an object so more landing params can be added later without new top-level props.

## Changes

* Hydration prefers `initialState.reportKey` over the server default; unknown keys warn and fall back
* Read once when config loads, never overrides user navigation
* Provider tests + `DeepLinkedReport` story

Consumers pass `{ reportKey }` at mount. Valid keys come from `/v1/businesses/{id}/reports/config`, which is per business.

## Blockers

None.

## How this has been tested?

* `npm test -- --run` (4 new tests), typecheck, lint
* Storybook: `DeepLinkedReport` lands on Balance Sheet, `Default` still on P&L

---

> [!NOTE]
> **Low Risk**
> Small public API addition with guarded hydration and tests; no auth or data-path changes.
> 
> **Overview**
> Adds optional **`initialState`** on **`UnifiedReports`** (and the store provider) so embedders can open a specific report instead of the server default—e.g. `{ reportKey: "BALANCE_SHEET" }` for deep links.
> 
> On first load after report config arrives, hydration picks **`initialState.reportKey`** when it exists in the business config; otherwise behavior is unchanged. Unknown keys **`console.warn`** and fall back to the default report. The requested key is applied only for initial hydration—it does not fight later user navigation.
> 
> Provider tests cover default hydration, requested key, invalid key fallback, and navigation after hydrate. Storybook adds **`DeepLinkedReport`** for Balance Sheet landing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f8ff0b10b462040c0fce87923e8368da76b8e3d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/providers/features/unifiedReports/UnifiedReportStore/UnifiedReportStoreProvider.tsx
+++ b/src/providers/features/unifiedReports/UnifiedReportStore/UnifiedReportStoreProvider.tsx
@@ -239,16 +239,11 @@ const findDefaultReport = (groups: ReadonlyArray<ReportGroup>): ReportConfig | n
   return firstReport
 }
 
-const hasReportWithKey = (groups: ReadonlyArray<ReportGroup>, key: string): boolean =>
-  groups.some(group => group.reports.some(report => report.key === key))
+const findReportByKey = (groups: ReadonlyArray<ReportGroup>, key: string): ReportConfig | null =>
+  groups.flatMap(({ reports }) => reports).find(report => report.key === key) ?? null
 
-const findReportByKey = (groups: ReadonlyArray<ReportGroup>, key: string): ReportConfig | null => {
-  for (const group of groups) {
-    const match = group.reports.find(report => report.key === key)
-    if (match) return match
-  }
-  return null
-}
+const hasReportWithKey = (groups: ReadonlyArray<ReportGroup>, key: string): boolean =>
+  findReportByKey(groups, key) != null
 
 const createUnifiedReportStore = (dateSelectionMode: DateSelectionMode) =>
   createStore<UnifiedReportStoreShape>(set => ({
--- a/src/providers/features/unifiedReports/UnifiedReportStore/UnifiedReportStoreProvider.tsx
+++ b/src/providers/features/unifiedReports/UnifiedReportStore/UnifiedReportStoreProvider.tsx
@@ -241,6 +241,14 @@ const findDefaultReport = (groups: ReadonlyArray<ReportGroup>): ReportConfig | n
 const hasReportWithKey = (groups: ReadonlyArray<ReportGroup>, key: string): boolean =>
   groups.some(group => group.reports.some(report => report.key === key))
 
+const findReportByKey = (groups: ReadonlyArray<ReportGroup>, key: string): ReportConfig | null => {
+  for (const group of groups) {
+    const match = group.reports.find(report => report.key === key)
+    if (match) return match
+  }
+  return null
+}
+
 const createUnifiedReportStore = (dateSelectionMode: DateSelectionMode) =>
   createStore<UnifiedReportStoreShape>(set => ({
     baseReport: null,
@@ -266,7 +274,7 @@ const createUnifiedReportStore = (dateSelectionMode: DateSelectionMode) =>
     },
   }))
 
-function useHydrateUnifiedReportStore(store: StoreApi<UnifiedReportStoreShape>) {
+function useHydrateUnifiedReportStore(store: StoreApi<UnifiedReportStoreShape>, defaultReportKey?: string) {
   const { data } = useGetReportConfig()
   const baseReport = useStore(store, state => state.baseReport)
   const setBaseReport = useStore(store, state => state.actions.setBaseReport)
@@ -275,9 +283,14 @@ function useHydrateUnifiedReportStore(store: StoreApi<UnifiedReportStoreShape>) 
     if (!data) return
     if (baseReport && hasReportWithKey(data, baseReport.key)) return
 
-    const defaultReport = findDefaultReport(data)
-    if (defaultReport) setBaseReport(defaultReport)
-  }, [data, baseReport, setBaseReport])
+    const requestedReport = defaultReportKey != null ? findReportByKey(data, defaultReportKey) : null
+    if (defaultReportKey != null && requestedReport == null) {
+      console.warn(`UnifiedReports: no report with key "${defaultReportKey}" in this business's report config; falling back to the default report.`)
+    }
+
+    const report = requestedReport ?? findDefaultReport(data)
+    if (report) setBaseReport(report)
+  }, [data, baseReport, setBaseReport, defaultReportKey])
 }
 
 function useSyncExternalDateSelectionMode(store: StoreApi<UnifiedReportStoreShape>, dateSelectionMode: DateSelectionMode) {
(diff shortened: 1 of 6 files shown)


sarahraines:
Reviewed against `src/testUtils/SKILL.md` and `.storybook/SKILL.md`. The behavior is right, and typecheck, lint, and the 4 new tests all pass on my end. Everything below is conventions, dead config, or comment trimming.

### 1. The `rerender()` in `does not re-assert the requested report over user navigation` is dead weight

`renderHookWithAuth`'s `rerender()` re-renders the hook with the same props, so the effect deps are unchanged and the effect doesn't re-run — the assertion right after `act` is what's actually doing the work. Confirmed by mutation: deleting the guard

```ts
if (baseReport && hasReportWithKey(data, baseReport.key)) return
```

still fails the test with those two lines removed (the `setBaseReport` inside `act` flushes the effect, which sees the new `baseReport` and snaps back to `BALANCE_SHEET`).

```diff
     act(() => result.current.setBaseReport(fixtureReport('TRIAL_BALANCE')))
     expect(result.current.baseReport?.key).toBe('TRIAL_BALANCE')
-
-    rerender()
-    expect(result.current.baseReport?.key).toBe('TRIAL_BALANCE')
```

Separately: the testing skill lists `act(() => …)` as an anti-pattern, and driving a store action + reading store state is the "asserting hook internals" case it warns about. If you want the user-navigation claim covered the way the repo does it, render `UnifiedReports` with `defaultState` and click a report in the sidebar. If that feels too heavy for this PR, the store-level test is fine — just drop the `rerender`.

### 2. The render helper should take `Partial<Props>` and be named after the subject

Per the skill: *"Name it after the subject … take `Partial<Props>` so each test states only the props it cares about."* Right now it redeclares a one-key subset of the provider's props inline, so a new provider prop needs a second edit here.

```diff
-const renderWithReportStore = (props?: { defaultState?: UnifiedReportsDefaultState }) =>
+const renderUnifiedReportStore = (props: Partial<ComponentProps<typeof UnifiedReportStoreProvider>> = {}) =>
   renderHookWithAuth(useBaseUnifiedReport, {
```

That also drops the `UnifiedReportsDefaultState` import from the test.

### 3. Fix the `renderHookWithAuth` footgun instead of commenting on it

```ts
// `renderHookWithAuth` spreads options over its own default wrapper, so this has to nest
// `LayerTestProvider` itself or auth never lands.
```

The comment is accurate, but it documents a sharp edge that every future caller will hit. Better to compose in the helper so a caller-supplied `wrapper` nests *inside* `LayerTestProvider` rather than replacing it — then this test just passes `UnifiedReportStoreProvider` as the wrapper and the comment goes away. All six existing callers pass no `wrapper`, so it's backwards compatible. If you'd rather not touch shared infra in this PR, cut the comment to one line.

### 4. `chromatic: { viewports: [1280] }` on `DeepLinkedReport` is dead config, and its comment is wrong

Two things:

- `views/**` stories aren't in the Chromatic build at all — `STORYBOOK_SCOPE=chromatic` scopes it to `ui/` and `blocks/` (see the header comment in `.github/workflows/chromatic.yml`), so the param never applies.
- The comment says the play function "needs a desktop width to assert against", but `scripts/check-stories-render.ts` always opens the page at `DOCS_SCREENSHOT_WIDTHS.desktop` (1280) regardless of story parameters. The play function is already at desktop width.

```diff
-// `defaultState` is read once at hydration, so changing the control on `Default` re-renders
-// without switching the report — only a fresh mount shows the behavior.
 export const DeepLinkedReport: Story = {
   args: { defaultState: { reportKey: 'BALANCE_SHEET' } },
-  // The report heading only renders in the desktop base header, so the play function
-  // needs a desktop width to assert against.
-  parameters: { chromatic: { viewports: [1280] } },
   play: async ({ canvasElement }) => {
```

The leading comment also just restates the `argTypes.description` two lines up. (`MegaMenuOpen` carries the same dead param — pre-existing, not yours to fix here, but worth not adding a second one.)

### 5. The `defaultState` control is a trap — document it, don't make it interactive

The description tells the reader the control doesn't do anything ("reload the story to see it apply"). A control that documents its own inertness is worse than no control. Keep it in `controls.include` so it still shows in the public docs args table, but turn off the control itself:

```diff
     defaultState: {
-      control: 'object',
-      description: 'Initial landing state (e.g. `{ reportKey: "BALANCE_SHEET" }`). Read once when the report configuration loads, so changing this control does not switch the report — reload the story to see it apply.',
+      control: false,
+      description: 'Initial landing state (e.g. `{ reportKey: "BALANCE_SHEET" }`), applied when the report configuration first loads. See the `DeepLinkedReport` story.',
     },
```

### 6. `findReportByKey` duplicates `hasReportWithKey`

Two traversals of the same shape, one of which is expressible in terms of the other:

```diff
-const hasReportWithKey = (groups: ReadonlyArray<ReportGroup>, key: string): boolean =>
-  groups.some(group => group.reports.some(report => report.key === key))
-
-const findReportByKey = (groups: ReadonlyArray<ReportGroup>, key: string): ReportConfig | null => {
-  for (const group of groups) {
-    const match = group.reports.find(report => report.key === key)
-    if (match) return match
-  }
-  return null
-}
+const findReportByKey = (groups: ReadonlyArray<ReportGroup>, key: string): ReportConfig | null =>
+  groups.flatMap(({ reports }) => reports).find(report => report.key === key) ?? null
+
+const hasReportWithKey = (groups: ReadonlyArray<ReportGroup>, key: string): boolean =>
+  findReportByKey(groups, key) != null
```

Typechecks and keeps all 4 tests green — I ran it.

### 7. Trim the JSDoc on the `defaultState` prop

Every other JSDoc block on a public view prop in this repo is a `@deprecated` marker; consumer-facing prop docs live in the Storybook `argTypes.description`, which this PR already writes. The last line (`Report switches are observable via the ReportsNavigated event`) isn't about this prop at all. Suggest one line, or none:

```diff
-  /**
-   * Initial landing state (e.g. `{ reportKey: 'PROFIT_AND_LOSS' }`) instead of the server default.
-   * Read when the report configuration loads — later changes do not switch the report
-   * (`defaultValue` semantics). Unknown report keys fall back to the default report with a console warning.
-   * Report switches are observable via the `ReportsNavigated` event.
-   */
+  /** Report to land on instead of the server default; applied once, when the report configuration loads. */
   defaultState?: UnifiedReportsDefaultState
```

### Not blocking, just noting

The JSDoc promises `defaultValue` semantics for *changes* to `defaultState`, but nothing tests that path — test 4 covers user navigation, not a consumer swapping the prop. Fine to leave uncovered given the early return also guards it, but it's the gap between what's documented and what's asserted.

sarahraines (approved):

src/providers/features/unifiedReports/UnifiedReportStore/UnifiedReportStoreProvider.tsx
```diff
@@ -288,11 +298,16 @@ function useSyncExternalDateSelectionMode(store: StoreApi<UnifiedReportStoreShap
 
 type UnifiedReportStoreProviderProps = {
   dateSelectionMode?: DateSelectionMode
+  defaultState?: UnifiedReportsDefaultState
```
sarahraines: Rename this prop to initialState
```suggestion
  initialState?: UnifiedReportsInitialState
```

PR #1749, opened 2026-08-12
[LAY-2615] fix(css): restore class names consumers style against

sarahraines:
## What broke

Between **v0.1.122** and today, a renaming pass moved most of the library from kebab-case to PascalCase BEM, and several components were restructured or replaced. Consumers style against our class names, so every selector written against the old names silently stopped matching — nothing in their build tells them.

Measured against two real consumer stylesheets, roughly **a fifth of the class names they target** no longer matched anything.

## Approach

Not a revert. Elements keep emitting the old names **alongside** the new ones, so our stylesheet stays on the current convention.

`createLegacyClassNames` binds a per-component map keyed by the *current* class name, so a map that goes stale fails typecheck rather than quietly dropping a name. Maps live inline next to the element they describe — only the data table needs a shared one.

```tsx
const legacyClassNames = createLegacyClassNames({
  Layer__IconBox: 'Layer__icon-box',
})

<span className={legacyClassNames('Layer__IconBox')}>
```

Keys that aren't class names (`variant:solid`, `state:fullWidth`) emit only their legacy value — for variants that are `data-*` attributes today.

**A feature's name never goes on a shared component.** `SubmitButton`, `MobileListItemStatusFooter` and `LedgerEntryDetails` are each rendered by several features that named their elements differently, so those names arrive from the caller. Putting them on the shared component would let a consumer's invoice rule style trips, or their bank-transaction rule style a confirmation modal.

## Also in here

- **Removed custom properties** restored in `legacy_variables.scss`. `--input-placeholder-color` was read by a bare, unscoped `::placeholder` rule that styled the whole host page; the knob is back, scoped to our roots.
- **A per-column hook** on the data table, so generated cells carry the names their hand-written `<td>`s used to.
- **`className` was being dropped** by `Header`, `Label`, `P`, `Button`, `Heading` and `ExpandableDataTable` — declared in props, never destructured, then overwritten by the hardcoded class. That silently swallowed several restorations, so it's fixed here.

## Tooling (`scripts/css-legacy`)

| | |
|---|---|
| `css:check-legacy-keys` | Fails on a map entry no call site reaches — it emits nothing while looking restored, which neither typecheck nor tests catch. Already caught two real bugs. |
| `css:rename` | Renames a class and seeds the legacy map in the same commit, so finishing the PascalCase migration can't repeat this. It does not rewrite the call site — that edit depends on how each element builds its class names, so the script lists the files still needing it and `css:check-legacy-keys` fails until they're done. |

`css:check-legacy-keys` is wired into `checks.yml` and blocking. A third script scanned every release since v0.1.122 for names that shipped and are gone today — that is how 83 of these were found, including names a v0.1.122-versus-today diff cannot see. It is not in here: its value was the one-time sweep, which this PR acts on, and keeping it meant carrying 328 lines of AST heuristics that mis-fire on ordinary refactors.

## Verification

313 tests pass; typecheck, lint, stylelint and all four size budgets clean. All 119 stories render without errors.

A class name is invisible in a rendered story, so three `*.scratch.stories.tsx` files read the emitted `class` attribute back out, printed it, and asserted on it in a `play` function — each paired with the state that should *not* carry the name. They have served their purpose and are removed from the branch.

Worth a Chromatic look: `Layer__table-cell-content` now sits on every table cell.

Root: #1749
Base: main
Next: #1765 — the checks that stop this from recurring

---

> [!NOTE]
> **Medium Risk**
> Wide DOM/className surface across the library affects every consumer that styles Layer components; mistakes could duplicate or omit legacy names, but changes are additive and guarded by a new CI check.
> 
> **Overview**
> Restores **pre-migration kebab-case and feature-specific class names** on DOM elements while keeping current PascalCase BEM for internal styles, so consumer CSS that targeted v0.1.122-era selectors works again without reverting the naming convention.
> 
> The change introduces **`createLegacyClassNames`** (and helpers like **`createLegacyFieldClassNames`**, shared table **`LEGACY_TABLE_CLASS_NAMES`**) wired through dozens of components: class strings are composed so **new and old names appear together**. Feature-specific legacy names are passed from callers into shared pieces (`MobileList`, `SummaryCard`, `LedgerEntryDetails`, tables via **`legacyClassNames` on column config**) so one feature’s selectors cannot leak onto another.
> 
> **Data tables** gain shared wrapper/header/cell/row legacy classes and per-column **`legacyClassNames`** metadata; **buttons** route variant/state through a central legacy map. Several components **fix `className` forwarding** (`Button`, `Header`, `ExpandableDataTable`, etc.) so restorations are not swallowed.
> 
> **CI** adds **`npm run css:check-legacy-keys`** (ts-morph scan for unreachable map keys). **`css:rename`** assists future renames. Date selection re-applies **mobile** legacy modifiers via **`useSizeClass`** where size used to drive class names.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5568b1e97875d442d817d967d5fa21ba79d0c019. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sarahraines (commented):
Let's do another review pass of all the threads and make sure they're addressed!

sarahraines (commented):
Testing if Augment rewakes.

sarahraines:
Do another review of this PR.

sarahraines:
Fix merge conflicts please. Also, clean up scratch stories and investigate/fix failing chromatic run.

sarahraines (commented):
You have a bug related to leading text input styles being affected by this change. Please see if you fix this by looking into the chromatic build on the Input variants story.

darrensapalo (approved):

PR #1752, opened 2026-08-13
[LAY-2612] test: report coverage in CI behind a manually raised floor

sarahraines:
## Scope

Measures test coverage on every PR and reports it, with a loose floor as a backstop. No new blocking gate beyond that floor.

- **`@vitest/coverage-v8`** + a `test:coverage` script. The existing `Vitest` check now runs it instead of `npm test -- --run`, so this adds no CI job — about 5s to an existing one.
- **Scope is all eligible source files**, not just tested ones. `include: src/**/*.{ts,tsx}` minus test infrastructure (msw, fixtures, testUtils, types, stories, tests) — 1,182 files, of which 183 currently have zero lines hit. An untested file counts against the number rather than being invisible, which is what lets the figure mean something.
- **Thresholds are a backstop, not a target.** They sit ~1pt below actual (`lines 30 / statements 29 / functions 16 / branches 11` against `31.08 / 29.87 / 16.69 / 11.97`). A global percentage moves when the denominator does — measured drift between two branches with an identical suite was 0.23pt — so a floor set at actual fails PRs that changed nothing about testing.
- **`autoUpdate` is off, deliberately.** It rewrites the thresholds on any improving run, which with ~20 open PRs means every one of them touching the same four lines. Raising the floor is instead an explicit act: `npm run coverage:floor` reports headroom, and `-- --write` raises any floor with 2pt or more of slack, targeting `floor(actual) - 1` so slack always lands in `[1pt, 2pt)`.
- **`testTimeout: 15_000`.** v8 instrumentation hits the slowest component tests hardest; `RecordBankTransactionModal` intermittently blew the 5s default under coverage. The comment records that test rather than an overhead percentage: the ratio depends on which phase you measure (+15% wall clock, +60% on the tests phase alone) and drifts as the suite grows.
- **Two reporters only: `text-summary` and `json-summary`.** The first prints in CI, the second is what `coverage:floor` reads. `html` and `lcov` are deliberately off: nothing consumes `lcov.info` (there is no coverage-upload step in `.github/workflows`), and the HTML reporter's generated JS and CSS were the sole reason `coverage/` needed an ESLint ignore and a stylelint ignore. Without them the directory is 386 KB instead of 37 MB and `npm run lint` passes with it on disk, so neither ignore is part of this PR. Pass `--coverage.reporter=html` on demand when hunting uncovered lines.

Also folds React Testing Library guidance into `src/testUtils/SKILL.md` — query priority, `userEvent` over `fireEvent`, async/`waitFor` discipline, an anti-pattern table, and the rule that every test file routes through a single module-scope render helper. Deduped against what was already there. The anti-pattern table deliberately restates rules the prose also covers: it is meant to be scanned mid-review, so it has to stand alone. The coverage-floor rationale, by contrast, lives in this file and only here — `vitest.config.ts` and `check-coverage-floor.ts` point at it instead of restating it.

## What this does not do

There's no per-PR diff-coverage gate, so a large untested feature can still land without failing anything — the floor only catches a real drop. That was a deliberate call, not an oversight.

## Verification

- `npm run lint`, `npm run typecheck`, `npm run test:coverage` all green (37 files / 284 tests).
- Confirmed the report covers all eligible files, not only tested ones: a single-test-file run still reports 1,182 files, matching the count on disk exactly.
- Confirmed dropping `html` and `lcov` changes nothing measured: totals are identical before and after, and `npm run lint` passes with `coverage/` present and both ignore entries deleted.
- Confirmed the floor blocks — the pre-existing thresholds from another branch failed here with exit 1 until re-baselined.
- Confirmed `coverage:floor --write` raises to the invariant rather than stepping: from a floor of 27 against 31.08% it goes straight to 30, and it refuses to lower a floor already above target. Re-verified after trimming the comments beside those four lines, since the script locates them by regex.

---

> [!NOTE]
> **Low Risk**
> CI and dev-tooling only; no production runtime or auth/data paths change beyond stricter test/coverage enforcement.
> 
> **Overview**
> CI’s **Vitest** job now runs **`npm run test:coverage`** (v8 via `@vitest/coverage-v8`) instead of a plain test run, with **global thresholds** in `vitest.config.ts` (~1pt below current totals, `autoUpdate: false`) so only a real coverage drop fails the build.
> 
> Coverage counts **all eligible `src/**` production files** (excluding msw, fixtures, testUtils, types, stories, and tests). **`npm run coverage:floor`** (`scripts/check-coverage-floor.ts`) reports headroom and can **`--write`** ratchet floors up deliberately. **`testTimeout`** is raised to **15s** for instrumentation overhead; **`coverage/`** is gitignored and excluded from ESLint/stylelint.
> 
> **`src/testUtils/SKILL.md`** is expanded with RTL guidance (queries, `userEvent`, async, module-scope render helpers, anti-patterns) and how to use the new coverage scripts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8e23189335df5e60f20610606adfc4341a3688c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/testUtils/SKILL.md
+++ b/src/testUtils/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: testing
-description: Testing conventions — vitest + Testing Library, LayerTestProvider as wrapper, MSW per-test mocks, form fillers, fixed clocks
+description: Testing conventions — vitest + Testing Library, query priority, LayerTestProvider as wrapper, MSW per-test mocks, form fillers, fixed clocks
 applies_to: src/**/*.test.ts, src/**/*.test.tsx, src/testUtils/**
 ---
 
@@ -10,7 +10,34 @@ applies_to: src/**/*.test.ts, src/**/*.test.tsx, src/testUtils/**
 Tests are colocated with the code (`CustomAccountForm.test.tsx` next to
 `CustomAccountForm.tsx`).
 
+The governing idea: **test what a user can perceive and do, never how the component achieves it.**
+Internal state, hook call counts, and prop plumbing are implementation details — asserting on them
+breaks under harmless refactors and passes when the wiring is genuinely broken.
+
 ```
+npm test                        # watch
+npm test -- --run               # single pass
+npm test -- --run path/to/file  # one file
+npm run test:coverage           # single pass + coverage (CI: Checks / Vitest)
+```
+
+Coverage floors live in `thresholds` in `vitest.config.ts`. They're a backstop against a real drop,
+not a target: a global percentage also moves when the denominator does, so they sit ~1pt below
+actual. Raising them is the ratchet, and it's deliberate rather than automatic — `autoUpdate` would
+have every concurrent PR rewriting the same four lines.
+
+```
+npm run coverage:floor            # headroom between actual and each floor
+npm run coverage:floor -- --write # raise any floor with 2pt+ of slack
+```
+
+Run it after landing a batch of tests, and commit the raise on its own.
+
+Note: the vitest glob excludes `.claude/**`, so agent worktrees are skipped by a bare
+`npm test`. Passing explicit file paths still works inside one.
+
+## Helpers
+
 Helpers live under `src/testUtils` (`@testUtils/*`), one directory per concern — put new ones
 where they belong instead of at the root:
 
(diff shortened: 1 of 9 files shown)


darrensapalo (approved):

sarahraines (commented):
Can you review this PR for brevity/cleanliness one more time?

src/testUtils/SKILL.md
```diff
@@ -229,18 +225,11 @@ will hang.
...
 | --- | --- |
-| `render(<Thing />)` inline in each `it()` | one module-scope `renderThing()` helper, used by every test |
 | `container.querySelector('.btn')` | `screen.getByRole('button', { name: /save/i })` |
-| `getByTestId('submit')` | `getByRole('button', { name: /submit/i })` |
-| `fireEvent.change(input, …)` | `await user.type(input, …)` |
-| `user.click(...)` un-awaited | `await user.click(...)` |
-| `await waitFor(() => screen.getByText('x'))` | `await screen.findByText('x')` |
-| `expect(queryByText('x')).toBeInTheDocument()` | `expect(getByText('x')).toBeInTheDocument()` |
-| 3 assertions in one `waitFor` | 1 inside, the rest after |
-| `user.click` inside `waitFor` | click first, then `waitFor` |
```
sarahraines: Wait, put back these antipattern examples.

PR #1760, opened 2026-08-17
[LAY-2607] perf: make the ESM build tree-shakeable

sarahraines:
## Scope

The ESM build shipped as a single 2.85 MB module, so a consumer's bundler had no boundaries to prune against and `sideEffects` had nothing to attach to — which is why [`6a1d91852`](https://github.com/Layer-Fi/layer-react/commit/6a1d91852) found adding the field byte-identical.

Importing one small component retained **505 modules / 539 kB** against the **57 modules** it actually needs, and pulled in all 28 runtime dependencies including `recharts`, `react-select`, `motion` and all three `@tanstack` packages.

This emits one file per source module and declares `sideEffects`. Both are load-bearing — per-module output alone changes nothing for consumers, verified below.

| size-limit (minified + brotli) | before | after | |
|---|---|---|---|
| full import (ESM) | 1.24 MB | 1.26 MB | +2% |
| single component (`GlobalMonthPicker`) | 550.28 kB | **105.51 kB** | −81% |
| single component (`BankTransactions`) | 596.62 kB | **453.14 kB** | −24% |
| stylesheet | 27.60 kB | 27.58 kB | — |

The full-import figure rises ~2%. That is the cost of real module boundaries and the trade that buys the 81%.

Of the remaining 105.51 kB for `GlobalMonthPicker`, only **9.0 kB is Layer code** — the rest is the 14 of 28 dependencies it genuinely needs. The floor is now third-party, not first-party.

### Changes

- **`vite.config.ts`** — `preserveModules: true` + `preserveModulesRoot: 'src'` on the ESM build. `dist/esm/` goes from one file to 1188 modules mirroring `src/`. `target` stays `es2016`; dropping an unimported module is a graph operation and does not depend on output syntax.
- **`package.json`** — adds `"sideEffects": ["**/*.css", "**/*.scss"]`. `main`, `module`, `types`, `style` and the `exports` map are untouched.
- **`src/index.tsx`** — the global stylesheet becomes the barrel's first import. `vite:css-post` crashes on a pure-SCSS entry under `preserveModules`, so the ordering contract moves from entry order to module execution order.
- **`vite/plugins/bundleCss.ts`** — collects whatever single stylesheet `cssCodeSplit: false` emits, since its name is no longer fixed.
- **`.size-limit.json`** — budgets retuned so the win cannot silently regress.
- **`consumer-fixtures/vite-esm-treeshake/`** — new fixture asserting a single-component import excludes nine unused heavy dependencies.

No consumer source changes. Import syntax, the `exports` map, `@layerfi/components/index.css` and types are all unchanged.

## Verification

`build:check`, `lint:package` (`publint` + `attw`, green across all four resolution modes), `test:types`, `typecheck`, `vitest` (43 files / 325 tests), and all 4 consumer fixtures.

**CSS is semantically unchanged.** Diffed against a pre-change baseline build: the rule set is identical — 1830 rules both sides, none added or removed. 101 rules shifted position, but the only genuine swap is two adjacent unrelated selectors and the rest is one contiguous block moving by a constant offset. No two rules sharing both a selector and a property changed relative order, and the global/component boundary sits at the same index in both files.

**The new fixture actually fails.** Removing `sideEffects` and doing a full rebuild makes all nine dependencies leak back in and the fixture errors out, with its bundle going 849 kB → 2,991 kB. That is also what establishes that `preserveModules` alone is not sufficient.

**`sideEffects` was audited, not assumed.** Every module-scope side effect was checked before declaring the field: `errorHandler.tsx`'s singleton only sets `onErrorCallback = undefined`; `i18n/init.ts` deliberately avoids the global `i18next` singleton; and `createResourceGlobalCacheActions` is a pure factory, so the ~19 discarded top-level calls are genuinely dead. All three bare imports esbuild drops from the built output were confirmed to have no top-level effects.

## Follow-ups (not in this PR)

- **CJS gets nothing from this.** `main` is still a single 3.12 MB `index.cjs`, so `require()` consumers are unaffected. Subpath exports via entry shims — the `react-aria-components` pattern, `"./*"` plus `"./private/*": null` — are the only thing that helps them, and are the highest-value next step if `main` has real consumers.
- `partner-accounting.jpg` is a 698 kB base64 data URI, 25% of the old bundle. Now isolated to `LandingPage`, so the other 30 exports no longer pay for it. Note `assetsInlineLimit` cannot fix this: Vite inlines all assets unconditionally in lib mode.
- `i18n/init.ts` statically pulls both `en-US` and `fr-CA` (~118 kB of JSON) for every consumer.
- `packageVersion.ts` inlines the entire `package.json`, including `scripts` and `devDependencies`, into the shipped bundle.

---

> [!NOTE]
> **Medium Risk**
> Changes the published ESM artifact layout and bundling contract (module graph + `sideEffects`); wrong config could break tree-shaking or CSS ordering for all ESM consumers, though consumer fixtures and size-limit mitigate regression risk.
> 
> **Overview**
> **Makes the published ESM build tree-shakeable** so importing one export no longer pulls the whole library graph. The ESM build now emits **one file per `src/` module** (`preserveModules`) and **`package.json` declares `sideEffects` for CSS/SCSS only**, giving consumer bundlers real prune boundaries. Single-component budgets drop sharply (e.g. `GlobalMonthPicker` ~550 kB → ~106 kB in size-limit); full ESM import rises slightly (~2%).
> 
> **Build/CSS wiring changes** to support that shape: global styles are the **first import on the barrel** (`src/index.tsx`) instead of a separate lib entry (pure-SCSS entry breaks under `preserveModules`), ESM uses **one merged stylesheet** (`cssCodeSplit: false` + updated `bundleCss`), and **size-limit thresholds** are retuned.
> 
> **Regression coverage**: new **`vite-esm-treeshake` consumer fixture** builds a minimal Vite app importing only `GlobalMonthPicker` and fails if heavy unused deps (`recharts`, `react-select`, `motion`, etc.) appear in the bundle. Storybook’s date mock is exposed as **`installSystemDateMock()`** so it stays reachable when side-effect-only imports are shaken.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d1b7dbd66ad34d3c03b984963621e3cc2383c536. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,3 +1,7 @@
+// Must stay the first import: global CSS has to precede component CSS in the generated
+// stylesheet so component styles win on equal specificity.
+import './styles/index.scss'
+
 /*
 ==========================================================
 ======================= Components =======================
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,5 +1,3 @@
-// Must stay the first import: global CSS has to precede component CSS in the generated
-// stylesheet so component styles win on equal specificity.
 import './styles/index.scss'
 
 /*
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -3,21 +3,21 @@
     "name": "full import (ESM)",
     "path": "dist/esm/index.mjs",
     "import": "*",
-    "limit": "1300 kB",
+    "limit": "1350 kB",
     "ignore": ["react", "react-dom"]
   },
   {
     "name": "single component, small (GlobalMonthPicker)",
     "path": "dist/esm/index.mjs",
     "import": "{ GlobalMonthPicker }",
-    "limit": "600 kB",
+    "limit": "120 kB",
     "ignore": ["react", "react-dom"]
   },
   {
     "name": "single component, large (BankTransactions)",
     "path": "dist/esm/index.mjs",
     "import": "{ BankTransactions }",
-    "limit": "650 kB",
+    "limit": "500 kB",
     "ignore": ["react", "react-dom"]
   },
   {
--- a/.storybook/mocks/systemDate.ts
+++ b/.storybook/mocks/systemDate.ts
@@ -1,22 +1,27 @@
 import { FIXTURE_YEAR } from '../../src/fixtures/constants/fixtureYear'
 
-const RealDate = Date
+// A bare `import './systemDate'` is a side-effect-only import, which `sideEffects` in
+// package.json (scoped to CSS/SCSS) makes eligible for tree-shaking under `preserveModules`.
+// Calling this from `preview.tsx` keeps the mock reachable regardless of that field.
+export function installSystemDateMock() {
+  const RealDate = Date
 
-// Clock ticks normally, but "now" is shifted to the last day of the fixture year.
-const offset = new RealDate(FIXTURE_YEAR, 11, 31, 12, 0, 0).getTime() - RealDate.now()
+  // Clock ticks normally, but "now" is shifted to the last day of the fixture year.
+  const offset = new RealDate(FIXTURE_YEAR, 11, 31, 12, 0, 0).getTime() - RealDate.now()
 
-// A Proxy rather than a subclass, so instances stay real Dates: with a subclass,
-// `value instanceof Date` is false for dates built before this module ran (e.g. the
-// fixture range constants), and query-param serialization silently falls back to
-// `String(date)` instead of an ISO date.
-globalThis.Date = new Proxy(RealDate, {
-  construct: (target, args: ConstructorParameters<DateConstructor>) =>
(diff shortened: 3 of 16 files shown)


sarahraines:
There's now an issue with the storybook build (date mocking), please investigate

src/index.tsx
```diff
@@ -1,3 +1,7 @@
+// Must stay the first import: global CSS has to precede component CSS in the generated
+// stylesheet so component styles win on equal specificity.
```
sarahraines: Remove this comment and any other useless ones regarding behavioral changes that reference old code.

PR #1800, opened 2026-09-10
[LAY-2717] feat: answer counterparty ask tasks in the Tasks component

AndrewAcomb:
## Description

Adds the answer flow for `ASK_ABOUT_COUNTERPARTY_FOR_PERIOD` tasks — the asks the agentic transaction categorizer creates per counterparty per period, rather than one task per transaction.

The owner sees the server-rendered question, then up to three suggested-account chips plus "Something else" (free text) and "A mix of the above" (per-transaction answers). A non-itemised answer goes on to a second step asking whether it should always apply, which is what creates the standing categorization rule. **Nothing is persisted until that second step completes** — one POST carries the whole answer, by design.

Backend contract: Layer-Fi/api#4992, with Layer-Fi/api#5033 adding the `legacy_tasks_only` opt-in these tasks need.

## Changes

**Schema** — `BusinessTaskSchema` becomes a `task_type`-discriminated union (`businessTasks/{base,legacy,counterpartyAsk}`). The legacy arm actively *excludes* the counterparty literal, so a malformed ask can't silently decode as a legacy task and POST to `/user-response`, which the API rejects with a 400. Every non-discriminator field on the ask arm is nullish-tolerant so one bad ask can't fail the whole periods response it rides in.

**Request body** — modelled as unions (`counterpartyAskResponse.ts`) so the four combinations the API 400s on are unrepresentable rather than guarded at the call site: account *and* free text on one answer, `always_this` alongside `transaction_responses`, neither, and blank text. The UI additionally blocks Continue until every linked transaction is answered.

**Read path** — the periods request now sends `legacy_tasks_only=false`. Without it the API excludes these tasks entirely and a period holding one renders "no pending tasks". The flag also feeds the period's derived status, so counts and the list agree.

**Cache invalidation** — a chip + "always" answer creates a rule, closes sibling asks in other periods, and applies retroactively across arbitrary months. The mutation therefore invalidates periods, bank transactions and P&L (the latter two only when an account was actually chosen, since free text categorizes nothing server-side), and reloads categorization rules only on the path that creates one.

**UI** — `TasksListItem` now branches on task type; the existing body moved to `LegacyTaskBody` unchanged. New `@ui/Chip` primitive built on radio-group semantics so a screen reader announces one choice of N. `BackButton` gains `isDisabled`.

## Blockers

None for merging, but three things are unresolved and worth knowing:

- **Copy is placeholder throughout** and not yet owned. Includes whether the second step lowercases the answer ("always be Business Meals?" vs the mockup's "business meals") — left verbatim because lowercasing translated strings misbehaves outside English.
- **`AccountStableName` wire shape unverified.** Uses the repo's existing `{type: 'StableName', stable_name: …}`; every suggestion on the test business is an `AccountId`, so the stable-name branch has never been exercised against a real response.
- **`resolved_by_task_id` can't name a period.** Only the uuid is serialized, so the copy says "Already answered" rather than "answered in November". That branch also can't render at all while `getUserVisibleTasks` drops `COMPLETED` tasks.

Chip layout is getting a revision — reviewers can skip nitpicking it.

## How this has been tested?

`typecheck`, `lint`, `i18n:check`, `msw:check-coverage`, `fixtures:check` all clean; **372 tests across 50 files**. Storybook builds.

Exercised end to end against **real production data** (business `c99cce33`, which has four live asks) via a local build behind ngrok — including a real "Yes, always" submit that created the rule, categorized retroactively, closed the sibling ask, and refreshed every affected cache.

Two bugs found that way and fixed, both worth a look since tests alone missed them:

- `transaction_responses` is the **link** table — one row per linked transaction from task creation, with null answers. Treating a non-empty array as "answered" hid the form on every open ask. My fixture had used `transactionResponses: []`, a state production never serves; it now derives unanswered rows from `transactions`, and reintroducing the bug fails 12 of 14 tests.
- The submit is slow enough under contention to click Back mid-flight, change the answer and resubmit — which the API correctly rejects as already-resolved, showing an error for a submission that succeeded. The confirm step now locks while in flight.

New tests cover the API's 400 conditions, the uniform-mix collapse, the `legacy_tasks_only` opt-in, the answered-state guard and the in-flight lock. Each was verified to fail when its fix is reverted.

---

> [!NOTE]
> **Medium Risk**
> New categorization and rule-creation paths POST to production APIs and invalidate bookkeeping/transaction caches; scope is large but guarded by schema unions, render filters, and extensive tests.
> 
> **Overview**
> **Adds an end-to-end UI for `ASK_ABOUT_COUNTERPARTY_FOR_PERIOD` bookkeeping tasks** so owners can answer agent-generated counterparty questions from the Tasks list.
> 
> `BusinessTask` is now a **`task_type`-discriminated union** (counterparty ask, legacy human, unknown/automated). Only renderable types reach the list; periods are fetched with **`legacy_tasks_only=false`** so these asks appear. **`TasksListItem`** delegates to new **`CounterpartyAskTaskBody`** (chip suggestions, free text, multi-transaction “mix,” optional **“always apply”** confirm before a single POST) or extracted **`LegacyTaskBody`** for existing flows.
> 
> Supporting work: **`@ui/Chip`** radio-group primitive, **`usePostCounterpartyAskResponse`** with periods/transactions/rules cache updates, response builders/schemas that forbid invalid API bodies, MSW + fixtures for monthly counterparty asks, Storybook story for no-suggestion asks, and broad unit/component tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4bc28ba101cf3fa439ee9c8eb8547902b4c6066. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/components/features/bookkeeping/TasksListItem/CounterpartyAskTaskBody.tsx
+++ b/src/components/features/bookkeeping/TasksListItem/CounterpartyAskTaskBody.tsx
@@ -1,0 +1,343 @@
+import { useCallback, useMemo, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+
+import { type CounterpartyAskResponse } from '@schemas/features/bookkeeping/businessTasks/counterpartyAskResponse'
+import { type CounterpartyAskTask } from '@schemas/features/bookkeeping/businessTasks/counterpartyAskTask'
+import { type UserVisibleTask } from '@utils/features/bookkeeping/bookkeepingTasksFilters'
+import {
+  buildAllSameCounterpartyAskResponse,
+  buildItemisedCounterpartyAskResponse,
+  collapseUniformCounterpartyAskAnswers,
+  countDistinctCounterpartyAskAnswers,
+  type CounterpartyAskAnswerValue,
+  getCounterpartyAskAnswerLabel,
+} from '@utils/features/bookkeeping/counterpartyAskAnswers'
+import { useLayerContext } from '@providers/global/LayerContext/LayerContext'
+import { usePostCounterpartyAskResponse } from '@api/businesses/[business-id]/tasks/[task-id]/counterparty-ask-response/post'
+import { BackButton } from '@ui/Button/BackButton'
+import { Button } from '@ui/Button/Button'
+import { Chip, ChipGroup } from '@ui/Chip/Chip'
+import { TextArea } from '@ui/Input/TextArea'
+import { HStack, VStack } from '@ui/Stack/Stack'
+import { P, Span } from '@ui/Typography/Text'
+import { CounterpartyAskTaskSummary } from '@features/bookkeeping/TasksListItem/CounterpartyAskTaskSummary'
+import {
+  CounterpartyAskTransactionRow,
+  OTHER_ANSWER_KEY,
+  toSuggestionAnswerKey,
+} from '@features/bookkeeping/TasksListItem/CounterpartyAskTransactionRow'
+
+import './counterpartyAskTaskBody.scss'
+
+const MIX_ANSWER_KEY = 'mix'
+
+type CounterpartyAskTaskBodyProps = {
+  task: UserVisibleTask & CounterpartyAskTask
+}
+
+export const CounterpartyAskTaskBody = ({ task }: CounterpartyAskTaskBodyProps) => {
+  const { t } = useTranslation()
+  const { addToast } = useLayerContext()
+  const { trigger: submitCounterpartyAskResponse, isMutating } = usePostCounterpartyAskResponse()
+
+  const [selectedKey, setSelectedKey] = useState<string | null>(null)
+  const [freeText, setFreeText] = useState('')
+  const [rowKeys, setRowKeys] = useState<Record<string, string>>({})
+  const [rowTexts, setRowTexts] = useState<Record<string, string>>({})
+  const [isConfirming, setIsConfirming] = useState(false)
+  const [sentAnswer, setSentAnswer] = useState<CounterpartyAskAnswerValue | null>(null)
+  const [sentDistinctCount, setSentDistinctCount] = useState<number | null>(null)
+
+  const { suggestions, transactions } = task
+  const totalTransactions = transactions.length
+  const isItemising = selectedKey === MIX_ANSWER_KEY
+  const isFreeText = selectedKey === OTHER_ANSWER_KEY
+
+  const resolveAnswer = useCallback(
+    (answerKey: string | undefined, text: string): CounterpartyAskAnswerValue | null => {
+      if (!answerKey) return null
+
+      if (answerKey === OTHER_ANSWER_KEY) {
+        const trimmed = text.trim()
+        return trimmed ? { kind: 'text', text: trimmed } : null
+      }
+
+      const suggestion = suggestions.find((_, index) => toSuggestionAnswerKey(index) === answerKey)
+
+      return suggestion ? { kind: 'account', account: suggestion } : null
+    },
+    [suggestions],
+  )
+
+  const answeredRows = useMemo(
+    () =>
+      transactions.flatMap((transaction) => {
+        const answer = resolveAnswer(rowKeys[transaction.id], rowTexts[transaction.id] ?? '')
+
+        return answer ? [{ transactionId: transaction.id, answer }] : []
+      }),
+    [transactions, rowKeys, rowTexts, resolveAnswer],
+  )
+
+  const isEveryRowAnswered = totalTransactions > 0 && answeredRows.length === totalTransactions
+
+  const singleAnswer = isItemising
+    ? (isEveryRowAnswered
+      ? collapseUniformCounterpartyAskAnswers(answeredRows.map(row => row.answer))
+      : null)
+    : resolveAnswer(selectedKey ?? undefined, freeText)
+
+  const canContinue = isItemising ? isEveryRowAnswered : singleAnswer !== null
+
+  const submit = useCallback(
+    async (response: CounterpartyAskResponse | null, onSuccess: () => void) => {
+      if (!response) return
+
+      try {
+        await submitCounterpartyAskResponse({ taskId: task.id, response })
+        onSuccess()
+      }
+      catch {
+        addToast({
+          content: t(
+            'bookkeeping:TasksListItem.CounterpartyAskTaskBody.error.submit_answer',
+            'We couldn’t save that answer. Please try again.',
+          ),
+          type: 'error',
+        })
+      }
+    },
+    [addToast, submitCounterpartyAskResponse, t, task.id],
+  )
+
+  const onContinue = useCallback(() => {
+    if (singleAnswer) {
+      setIsConfirming(true)
+      return
+    }
+
+    void submit(buildItemisedCounterpartyAskResponse(answeredRows), () => {
+      setSentDistinctCount(countDistinctCounterpartyAskAnswers(answeredRows.map(row => row.answer)))
+    })
+  }, [answeredRows, singleAnswer, submit])
+
+  const onConfirm = useCallback(
+    (alwaysThis: boolean) => {
+      if (!singleAnswer) return
+
+      void submit(buildAllSameCounterpartyAskResponse(singleAnswer, alwaysThis), () => {
+        setSentAnswer(singleAnswer)
+      })
+    },
+    [singleAnswer, submit],
+  )
+
+  if (sentAnswer) {
+    return (
+      <CounterpartyAskTaskSummary
+        title={t('bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.answer_sent', 'Thanks — sent')}
+        detail={t(
+          'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.receipt_all_same',
+          'All {{total}} · {{answer}}',
+          { total: totalTransactions, answer: getCounterpartyAskAnswerLabel(sentAnswer) },
+        )}
+        note={t(
+          'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.receipt_note_all_same',
+          'We’ll use this for your books.',
+        )}
+      />
+    )
+  }
+
+  if (sentDistinctCount !== null) {
+    return (
+      <CounterpartyAskTaskSummary
+        title={t('bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.answer_sent', 'Thanks — sent')}
+        detail={t(
+          'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.receipt_itemised',
+          '{{total}} transactions · {{distinct}} different answers',
+          { total: totalTransactions, distinct: sentDistinctCount },
+        )}
+        note={t(
+          'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.receipt_note_itemised',
+          'Your bookkeeper takes it from here.',
+        )}
+      />
+    )
+  }
+
+  if (task.resolvedByTaskId) {
+    return (
+      <CounterpartyAskTaskSummary
+        title={t(
+          'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.answered_elsewhere',
+          'Already answered',
+        )}
+        detail={t(
+          'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.answered_elsewhere_detail',
+          'You answered this for every period, so we’ve applied it here too.',
+        )}
+      />
+    )
+  }
+
+  const storedAnswer: CounterpartyAskAnswerValue | null = task.responseAccount
+    ? { kind: 'account', account: task.responseAccount }
+    : (task.userResponse ? { kind: 'text', text: task.userResponse } : null)
+
+  if (storedAnswer) {
+    return (
+      <CounterpartyAskTaskSummary
+        title={t('bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.answered', 'Answered')}
+        detail={t(
+          'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.receipt_all_same',
+          'All {{total}} · {{answer}}',
+          { total: totalTransactions, answer: getCounterpartyAskAnswerLabel(storedAnswer) },
+        )}
+        note={task.alwaysThis
+          ? t(
+            'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.answered_always',
+            'We’ll use this every time from now on.',
+          )
+          : undefined}
+      />
+    )
+  }
+
+  const answeredTransactions = task.transactionResponses.filter(
+    response => Boolean(response.userResponse) || Boolean(response.responseAccount),
+  )
+
+  if (answeredTransactions.length > 0) {
+    return (
+      <CounterpartyAskTaskSummary
+        title={t('bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.answered', 'Answered')}
+        detail={t(
+          'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.answered_itemised',
+          'You answered {{answered}} of {{total}} individually.',
+          { answered: answeredTransactions.length, total: totalTransactions },
+        )}
+      />
+    )
+  }
+
+  if (isConfirming && singleAnswer) {
+    const answerLabel = getCounterpartyAskAnswerLabel(singleAnswer)
+    const counterpartyName = task.counterparty?.name
+
+    return (
+      <VStack gap='md' pb='sm'>
+        <HStack>
+          <BackButton onPress={() => setIsConfirming(false)} isDisabled={isMutating} />
+        </HStack>
+        <P weight='bold'>
+          {counterpartyName
+            ? t(
+              'bookkeeping:TasksListItem.CounterpartyAskTaskBody.prompt.always_this_for_counterparty',
+              'Will {{counterparty}} purchases always be {{answer}}?',
+              { counterparty: counterpartyName, answer: answerLabel },
+            )
+            : t(
+              'bookkeeping:TasksListItem.CounterpartyAskTaskBody.prompt.always_this',
+              'Should we always use {{answer}} for these?',
+              { answer: answerLabel },
+            )}
+        </P>
+        <HStack gap='2xs'>
+          <Button isDisabled={isMutating} onPress={() => onConfirm(true)}>
+            {t('bookkeeping:TasksListItem.CounterpartyAskTaskBody.action.yes_always', 'Yes, always')}
+          </Button>
+          <Button variant='outlined' isDisabled={isMutating} onPress={() => onConfirm(false)}>
+            {t('bookkeeping:TasksListItem.CounterpartyAskTaskBody.action.no_ask_again', 'No, ask me again')}
+          </Button>
+        </HStack>
+      </VStack>
+    )
+  }
+
+  return (
+    <VStack gap='sm' pb='sm'>
+      <P size='sm' variant='inherit'>{task.question}</P>
+      <ChipGroup
+        ariaLabel={t(
+          'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.answer',
+          'What these were for',
+        )}
+        value={selectedKey}
+        onChange={setSelectedKey}
+        wrap
+      >
+        {suggestions.map((suggestion, index) => (
+          <Chip key={toSuggestionAnswerKey(index)} value={toSuggestionAnswerKey(index)}>
+            {suggestion.name}
+          </Chip>
+        ))}
+        <Chip value={OTHER_ANSWER_KEY}>
+          {t('bookkeeping:TasksListItem.CounterpartyAskTaskBody.action.something_else', 'Something else')}
+        </Chip>
+        {totalTransactions > 1
+          ? (
+            <HStack className='Layer__CounterpartyAskTask__MixRow' justify='center'>
+              <Chip value={MIX_ANSWER_KEY}>
+                {t('bookkeeping:TasksListItem.CounterpartyAskTaskBody.action.a_mix_of_the_above', 'A mix of the above')}
+              </Chip>
+            </HStack>
+          )
+          : null}
+      </ChipGroup>
+      {isFreeText
+        ? (
+          <VStack className='Layer__CounterpartyAskTask__FreeText'>
+            <TextArea
+              value={freeText}
+              placeholder={t(
+                'bookkeeping:TasksListItem.CounterpartyAskTaskBody.placeholder.what_were_these_for',
+                'What were these purchases for?',
+              )}
+              onChange={event => setFreeText(event.target.value)}
+            />
+          </VStack>
+        )
+        : null}
+      {isItemising
+        ? (
+          <VStack gap='2xs'>
+            <VStack className='Layer__CounterpartyAskTask__Rows'>
+              {transactions.map(transaction => (
+                <CounterpartyAskTransactionRow
+                  key={transaction.id}
+                  transaction={transaction}
+                  suggestions={suggestions}
+                  selectedKey={rowKeys[transaction.id] ?? null}
+                  text={rowTexts[transaction.id] ?? ''}
+                  onSelect={answerKey =>
+                    setRowKeys(current => ({ ...current, [transaction.id]: answerKey }))}
+                  onChangeText={text =>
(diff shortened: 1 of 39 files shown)


src/schemas/features/bookkeeping/businessTasks/counterpartyAskTask.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export type CounterpartyAskTransactionResponse = typeof CounterpartyAskTransactionResponseSchema.Type
+
+export const CounterpartyAskTaskSchema = Schema.extend(
+  BaseBusinessTaskSchema,
+  Schema.Struct({
+    taskType: pipe(
+      Schema.propertySignature(Schema.Literal(COUNTERPARTY_ASK_TASK_TYPE)),
+      Schema.fromKey('task_type'),
+    ),
```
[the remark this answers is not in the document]
AndrewAcomb: Fixed in e5f5cf6; regression test in 3a7bdb8.

src/components/features/bookkeeping/TasksListItem/CounterpartyAskTaskBody.tsx
```diff
@@ -0,0 +1,343 @@
...
+import './counterpartyAskTaskBody.scss'
+
+const MIX_ANSWER_KEY = 'mix'
+
+type CounterpartyAskTaskBodyProps = {
+  task: UserVisibleTask & CounterpartyAskTask
+}
+
+export const CounterpartyAskTaskBody = ({ task }: CounterpartyAskTaskBodyProps) => {
+  const { t } = useTranslation()
```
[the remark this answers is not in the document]
AndrewAcomb: Fixed in e5f5cf6, with tests.

AndrewAcomb:
augment review

AndrewAcomb:
Reviewed from the backend side only — contract correctness and forward compatibility. Other agents are covering the holistic review, so I stayed out of component structure, styling and test coverage.

Four inline comments. Ranked:

1. **The task union has no fallback** and the backend can already emit shapes it rejects (`Automated_Task` and `Automated_Transaction_Review_Task` omit `question`, and `deprecatedTasks` does not filter them). Pre-existing fragility rather than a regression, but this PR widens the exposure by defaulting the periods hook to `legacy_tasks_only=false`. This is the one I would not merge without at least confirming those types don't occur for bookkeeping businesses.
2. Itemised submissions must cover every linked transaction or the server 400s, and the schema can't express that — worth confirming the submit gate handles accretion, not just the count at load.
3. `legacy_tasks_only=false` means "include every agent-created type", which is broader than it reads.
4. Minor over-firing of the bank-transaction invalidation on mixed itemised answers.

The submit schema is well matched to the server: the itemised branch structurally cannot carry `always_this`, and the rule-invalidation guard mirrors the backend condition exactly. Those are the two places I expected to find drift and didn't.

src/schemas/features/bookkeeping/businessTask.ts
```diff
@@ -1,65 +1,27 @@
...
+import {
+  type CounterpartyAskTask,
+  CounterpartyAskTaskSchema,
+} from '@schemas/features/bookkeeping/businessTasks/counterpartyAskTask'
+import {
+  type LegacyBusinessTask,
+  LegacyBusinessTaskSchema,
+} from '@schemas/features/bookkeeping/businessTasks/legacyBusinessTask'
+
+export const BusinessTaskSchema = Schema.Union(
```
AndrewAcomb, line 14: Backend perspective: this union has no fallback member, and the backend can already emit task shapes that neither member accepts.

`BaseBusinessTaskSchema` requires `question`, and `LegacyBusinessTaskSchema` additionally requires the `user_response` and `user_response_type` keys. Two backend types serialize none of the three:

- `Automated_Task` (`AUTOMATED_RULE_SUGGESTION`) — carries `suggestion`, no `question`
- `Automated_Transaction_Review_Task` — no `question`, no `user_response_type`

Neither is `fromLlmTransactionCategorizer`, so `BookkeepingPeriod.deprecatedTasks` does **not** filter them out — they come through the periods endpoint regardless of `legacy_tasks_only`. One such task fails both union members and takes the whole periods decode down with it.

This fragility predates the PR (the old single-Struct schema required the same fields), so it isn't a regression. But it gets more exposed here: the periods hook now defaults to `legacy_tasks_only=false`, which means "include every agent-created type", not "include counterparty asks". When I add the next agent task type on the backend it will land in this response and break the list rather than being skipped.

Suggest a permissive third member (or making the base fields optional) so an unrecognised shape degrades instead of throwing. Worth also checking whether those two automated types actually occur for bookkeeping businesses — if they do, this is live today rather than hypothetical.

src/hooks/api/businesses/[business-id]/bookkeeping/periods/get.ts
```diff
@@ -50,6 +56,7 @@ const useBookkeepingPeriodsQuery = createQueryHook({
   tags: [BOOKKEEPING_TAG_KEY, BOOKKEEPING_PERIODS_TAG_KEY],
   request: getBookkeepingPeriods,
   schema: BookkeepingPeriodsResponseSchema,
+  keyDefaults: { legacyTasksOnly: false },
```
Comment (AndrewAcomb) [inline src/hooks/api/businesses/[business-id]/bookkeeping/periods/get.ts:63]: Worth a comment here on what this flag actually means on the backend, because the name undersells it.

`legacy_tasks_only=false` removes the filter entirely — it is "include every `fromLlmTransactionCategorizer` type", not "include counterparty asks". Today that set is exactly one type, so this is correct. As soon as I add another agent-created task type it will start arriving here too, which is the forward-compat concern I left on the union in `businessTask.ts`.

I picked the name to match the existing parameter on `GET /v1/businesses/{id}/tasks` rather than inventing a second convention, so the semantics are inherited rather than chosen. Flagging so the coupling is visible at the call site.

src/utils/features/bookkeeping/counterpartyAskAnswers.ts
```diff
@@ -0,0 +1,80 @@
...
+  answer: CounterpartyAskAnswerValue,
+  alwaysThis: boolean,
+): CounterpartyAskResponse => ({ ...toAnswer(answer), alwaysThis })
+
+export type CounterpartyAskTransactionAnswerEntry = {
+  transactionId: string
+  answer: CounterpartyAskAnswerValue
+}
+
+export const buildItemisedCounterpartyAskResponse = (
```
AndrewAcomb, line 67: The backend rejects an itemised submission that omits **any** linked transaction — `applyItemisedCounterpartyAskResponse` 400s with "An itemised response must answer every linked transaction". `NonEmptyArray` on the schema catches the empty case, but full coverage cannot be expressed there, so it has to be enforced by whatever gates the submit button.

Worth confirming the confirm step is disabled until every linked transaction has an answer, including any that arrived via accretion after the task was first rendered. That second part is the easy one to miss: the task can gain transactions between load and submit, and the count you validated against is the stale one.

src/hooks/api/businesses/[business-id]/tasks/[task-id]/counterparty-ask-response/post.ts
```diff
@@ -0,0 +1,63 @@
...
+  swrOptions: { throwOnError: true },
+  useOnTriggerSuccess: () => {
+    const { invalidate: invalidateBookkeepingPeriods } = useBookkeepingPeriodsGlobalCacheActions()
+    const { forceReload: forceReloadCategorizationRules } = useCategorizationRulesGlobalCacheActions()
+    const onBankTransactionChange = useBankTransactionTriggerSuccess()
+
+    return (task) => {
+      void invalidateBookkeepingPeriods()
+
+      const categorized = Boolean(task.responseAccount)
```
Comment (AndrewAcomb) [inline src/hooks/api/businesses/[business-id]/tasks/[task-id]/counterparty-ask-response/post.ts]: Minor, and only a wasted refetch rather than a bug: this over-fires on a mixed itemised answer.

On the backend, one free-text row disables synchronous categorization for the **whole** task — `applyItemisedCounterpartyAskResponse` still writes the chosen account onto each chip row's link, then returns no categorizations if any row was free text. So the response comes back with `response_account` populated on some `transaction_responses` while nothing was actually categorized, and `some(response => response.responseAccount)` is true.

If you want it exact, the condition is "every itemised row has an account":

```ts
const categorized = Boolean(task.responseAccount)
  || (task.transactionResponses.length > 0
    && task.transactionResponses.every(response => response.responseAccount))
```

The `alwaysThis && responseAccount` guard below it is exactly right, for what it's worth — a rule is only created for a chip answer plus always, never for free text and never from the itemised path.

src/components/features/bookkeeping/TasksListItem/CounterpartyAskTaskBody.tsx
```diff
@@ -0,0 +1,350 @@
...
+        )}
+        note={t(
+          'bookkeeping:TasksListItem.CounterpartyAskTaskBody.label.receipt_note_itemised',
+          'Your bookkeeper takes it from here.',
+        )}
+      />
+    )
+  }
+
+  if (task.resolvedByTaskId) {
```
AndrewAcomb: **Answered state is derived from payload fields, not `task.status`.**

This branch (and the `responseAccount` / `userResponse` / answered-rows branches below) decides "already answered" without reading `status`, while every sibling keys off `status` alone: `isIncompleteTask`/`isCompletedTask`, `getIconForTask`, the `TasksList` sort and `defaultOpen`, the `TasksMonthSelector` badge, and `useBookkeepingYearsStatus`.

So a `TODO` task with `resolved_by_task_id` set (which is exactly what the "resolved by another period" test builds on the fixture's default Todo status) shows the todo icon, sorts first, auto-expands, counts as pending in the month badge, yet the body says "Already answered" with no Continue button. The user can never clear the count.

If the backend instead marks these `COMPLETED`, `getUserVisibleTasks` drops them and this branch is dead code. Either way it'd be good to pick one source of truth, probably `status`, and let the payload fields only choose which summary copy to show.

src/schemas/features/bookkeeping/businessTask.ts
AndrewAcomb, line 14: **Strict two-arm union now guards the entire periods response.**

The periods GET now sends `legacy_tasks_only=false`, so every non-legacy task the backend returns has to decode as one of these two arms. The counterparty arm still requires `counterparty.mccs`, `transactions[].date` as `Schema.Date`, `transactions[].direction` and a UUID `id`; the legacy arm rejects the ask literal and requires a non-empty `user_response_type`.

One ask missing `mccs`, one bad date, or any new `task_type` without `user_response_type` rejects the whole `/bookkeeping/periods` body in `createKeyedFetcher`. Then `useGetBookkeepingPeriods.data` is undefined for every month and Tasks / month selector / year status render nothing, including the legacy tasks that worked before this PR.

The inverse is also possible: an unknown non-legacy type that happens to carry `user_response_type` decodes as legacy and renders the free-response form.

Worth considering a per-element lenient decode (drop the element, keep the period) or a catch-all arm that `getUserVisibleTasks` filters out.

src/components/features/bookkeeping/TasksListItem/CounterpartyAskTaskBody.tsx
```diff
@@ -0,0 +1,350 @@
...
+          : undefined}
+      />
+    )
+  }
+
+  const answeredTransactions = task.transactionResponses.filter(
+    response => Boolean(response.userResponse) || Boolean(response.responseAccount),
+  )
+
+  if (answeredTransactions.length > 0) {
```
AndrewAcomb: **Any answered row is treated as terminal.**

One answered `transaction_responses` row hides the form and the Continue button even when `status` is `TODO` and other rows are still unanswered. The per-row schema makes partial state representable (`user_response` / `response_account` are `NullishOr` per row), and the test at `CounterpartyAskTaskBody.test.tsx:252` codifies exactly this partial-on-Todo shape.

If partial state can come back from the API (bookkeeper tooling, another client), the task sits in the incomplete list with no way to submit the rest. Ties into the `status` comment above: gating the summary view on `status` would fix both.

src/hooks/api/businesses/[business-id]/tasks/[task-id]/counterparty-ask-response/post.ts
Comment (AndrewAcomb) [inline src/hooks/api/businesses/[business-id]/tasks/[task-id]/counterparty-ask-response/post.ts]: **Invalidation is gated on the response echo.**

`onBankTransactionChange()` and `forceReloadCategorizationRules()` only fire if the server echoes `response_account` / `always_this`. Both are `NullishOr` / default false in the schema, so a 200 with those omitted (async categorization, StableName resolving differently, contract drift) silently skips the Bank Transactions, P&L and Rules refresh.

Meanwhile the fix commit derives the same fact from the *request* in the component (`wasCategorized`), so one flow now has two sources of truth. Every other consumer of `useBankTransactionTriggerSuccess` invalidates unconditionally. Suggest doing the same here and dropping the echo check.

(The MSW handler always synthesizes `response_account` from the request, so no test can observe a null echo.)

src/components/ui/Chip/Chip.tsx
```diff
@@ -0,0 +1,70 @@
...
+  ...restProps
+}: ChipGroupProps<T>) {
+  const dataProperties = toDataProperties({ wrap })
+
+  return (
+    <ReactAriaRadioGroup
+      {...restProps}
+      {...dataProperties}
+      aria-label={ariaLabel}
+      value={value ?? null}
```
AndrewAcomb: **`ChipGroup` can never be uncontrolled.**

`value ?? null` means react-stately's `useControlledState` always sees `value !== undefined` and pins the group to controlled mode, and `defaultValue` is `Omit`-ed from the props. The two stories in `Chip.stories.tsx` that omit `value` fire `onChange` but never show a selection.

Both production consumers pass a controlled value, so this is Storybook-only today. Either pass `value` through as-is (letting `undefined` mean uncontrolled) or expose `defaultValue`.

src/components/features/bookkeeping/TasksListItem/CounterpartyAskTransactionRow.tsx
```diff
@@ -0,0 +1,84 @@
...
+  onSelect,
+  onChangeText,
+}: CounterpartyAskTransactionRowProps) => {
+  const { t } = useTranslation()
+  const { formatDate } = useIntlFormatter()
+
+  return (
+    <VStack className='Layer__CounterpartyAskTask__Row' gap='2xs' pb='2xs' pi='2xs'>
+      <HStack align='baseline' gap='xs'>
+        <MoneySpan size='sm' weight='bold' amount={transaction.amount} />
```
AndrewAcomb: **Row ignores `transaction.direction`.**

`MinimalBankTransactionSchema` is an unsigned amount plus direction, and every other renderer of it (`AffectedTransactionsTable`, `BankTransactionsTable`, `BankTransactionsListItem`) passes `displayPlusSign={direction === Credit}`. Here a Costco refund renders identically to a purchase, so someone picking "a mix of the above" can't tell inflows from outflows.

Fixtures and the component test are all `DEBIT`, so the credit path is untested.

src/fixtures/bookkeeping/counterpartyAskTasks.ts
```diff
@@ -0,0 +1,168 @@
...
+import {
+  type CounterpartyAskAccount,
+  type CounterpartyAskTask,
+} from '@schemas/features/bookkeeping/businessTasks/counterpartyAskTask'
+
+import { FIXTURE_YEAR } from '@fixtures/constants/fixtureYear'
+import { createFixtureFactory } from '@fixtures/utils/createFixtureFactory'
+
+const RETAIL_SUGGESTIONS: readonly CounterpartyAskAccount[] = [
+  { accountIdentifier: makeAccountId('00000000-0000-4000-8000-000000000801'), name: 'Office Expenses' },
```
AndrewAcomb: **Hand-rolled account ids collide with other fixtures.**

`...0801` is `Cash` in `src/fixtures/chartOfAccounts/mocks.ts` and `...0802` is `Sales` in `src/fixtures/ledgerEntries/mocks.ts`. Neither is in the mocked chart of accounts under these names. `src/fixtures/bankTransactions/constants.ts` already exports `bankTransactionCategories.officeExpenses` / `meals` / `rent` resolved from `chartOfAccounts.gen` with the ids the mocked `/categories` endpoint actually serves; reusing those keeps stories and tests consistent when they look an account up by id after an "always" answer.

src/fixtures/bookkeeping/mocks.ts
```diff
@@ -101,7 +100,7 @@ const monthsBeforeCurrent = (year: number, month: number) => {
 /** Past months without open tasks have closed books, so they carry no uncategorized activity. */
 export const hasCompletedBooks = (year: number, month: number) => {
   const monthsAgo = monthsBeforeCurrent(year, month)
-  return monthsAgo > 0 && openTaskCountFor(monthsAgo) === 0
+  return monthsAgo > 0 && openTaskCountFor(monthsAgo) + counterpartyAskCountFor(year, month) === 0
```
AndrewAcomb: **Side effect on P&L fixtures and docs screenshots.**

`makeBookkeepingPeriods` now seeds a TODO ask for `FIXTURE_YEAR` months 7 through 11, and `hasCompletedBooks` counts them. Storybook pins now to Dec 31 of `FIXTURE_YEAR`, so August and October (months-ago 4 and 2, zero legacy open tasks) flip from `CLOSED_COMPLETE` to `IN_PROGRESS_AWAITING_CUSTOMER`. The shared P&L fixture (`src/fixtures/profitAndLoss/mocks.ts:39`) then stops zeroing uncategorized inflows/outflows for those months.

Net effect: the Tasks Default story goes from 3 to 4 open tasks (probably intended), and the P&L docs-screenshot stories pinned to September show non-zero uncategorized bars for August and October where the books were previously clean. Chromatic baselines and docs screenshots move even though no P&L code changed. If that's unintended, seed the asks only for months that already have open legacy tasks, or exclude them from `hasCompletedBooks`.

src/msw/api/businesses/[business-id]/tasks/[task-id]/upload/delete/post.ts
```diff
@@ -1,10 +1,7 @@
 import { Schema } from 'effect'
 
-import {
-  type BusinessTask,
-  BusinessTaskSchema,
-  BusinessTaskStatus,
-} from '@schemas/features/bookkeeping/businessTask'
+import { type BusinessTask, BusinessTaskSchema } from '@schemas/features/bookkeeping/businessTask'
```
Comment (AndrewAcomb) [inline src/msw/api/businesses/[business-id]/tasks/[task-id]/upload/delete/post.ts]: **Only `upload/post.ts` got the `isLegacyBusinessTask` guard.**

This handler (the `patchTaskInStore` call at line 23), `upload/update-description`, and `completeTaskInStore` in `store.ts` still spread `documents: null` / `userResponse` / status onto whatever task matches the id, including counterparty asks now in the periods store. It typechecks because `patchTaskInStore` is typed over the union and `userResponse` exists on the ask arm.

A test or story that POSTs `/tasks/<ask-id>/user-response` would mark the ask `USER_MARKED_COMPLETED` and `CounterpartyAskTaskBody` would render it as an all-same answer, while the real API 400s that call. Latent today, but worth guarding the siblings the same way.

AndrewAcomb (commented):
Review notes below. Findings 1 and 2 are the ones I'd want resolved before merge; the rest are smaller.

One thing I didn't leave inline: the POST reply is strictly decoded with `throwOnError: true`, so a write that succeeds server-side but echoes an unexpected shape toasts an error and leaves the form up, and the retry hits the already-resolved 400. That matches the `src/hooks/api` convention so I'm treating it as an accepted trade-off, just flagging that the consequence is worse here than for the legacy siblings.

src/schemas/features/bookkeeping/businessTask.ts
AndrewAcomb: Catch-all arm added in 5e7d53a.

AndrewAcomb: Catch-all arm added in 5e7d53a.

src/hooks/api/businesses/[business-id]/tasks/[task-id]/counterparty-ask-response/post.ts
Comment (AndrewAcomb) [reply src/hooks/api/businesses/[business-id]/tasks/[task-id]/counterparty-ask-response/post.ts]: Fixed in fdf4c66; now requires every row.

Comment (AndrewAcomb) [reply src/hooks/api/businesses/[business-id]/tasks/[task-id]/counterparty-ask-response/post.ts]: Now invalidates unconditionally in fdf4c66.

src/components/features/bookkeeping/TasksListItem/CounterpartyAskTaskBody.tsx
AndrewAcomb: Fixed in fdf4c66; gated on status.

src/components/features/bookkeeping/TasksListItem/CounterpartyAskTaskBody.tsx
AndrewAcomb: Fixed in fdf4c66; gated on status.

src/components/ui/Chip/Chip.tsx
AndrewAcomb: Fixed in fdf4c66; undefined now means uncontrolled.

src/components/features/bookkeeping/TasksListItem/CounterpartyAskTransactionRow.tsx
AndrewAcomb: Fixed in fdf4c66.

src/fixtures/bookkeeping/counterpartyAskTasks.ts
AndrewAcomb: Fixed in fdf4c66; uses shared category fixtures.

src/msw/api/businesses/[business-id]/tasks/[task-id]/upload/delete/post.ts
Comment (AndrewAcomb) [reply src/msw/api/businesses/[business-id]/tasks/[task-id]/upload/delete/post.ts]: Fixed in fdf4c66; siblings guarded.

src/schemas/features/bookkeeping/businessTask.ts
```diff
@@ -1,65 +1,37 @@
...
+// unrecognised task_type this union can't fully model) never does.
+export const isLegacyBusinessTask = <T extends Pick<BusinessTask, 'taskType'>>(
+  task: T,
+): task is T & LegacyBusinessTask =>
+  !isCounterpartyAskTask(task) && 'userResponseType' in task
+
+export const isRenderableBusinessTask = <T extends BusinessTask>(
+  task: T,
+): task is T & (CounterpartyAskTask | LegacyBusinessTask) =>
+  isCounterpartyAskTask(task) || isLegacyBusinessTask(task)
```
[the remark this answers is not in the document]
AndrewAcomb: Fixed in eccc184; guard now requires the ask shape.

src/fixtures/bookkeeping/mocks.ts
AndrewAcomb: Fixed in eccc184; asks no longer affect hasCompletedBooks.

AndrewAcomb:
augment review

AndrewAcomb:
cursor review

src/components/features/bookkeeping/TasksListItem/CounterpartyAskTaskBody.tsx
```diff
@@ -0,0 +1,353 @@
...
+          </Button>
+        </HStack>
+      </VStack>
+    )
+  }
+
+  return (
+    <VStack gap='sm' pb='sm'>
+      <P size='sm' variant='inherit'>{task.question}</P>
+      <ChipGroup
```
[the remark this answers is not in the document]
AndrewAcomb: Valid against the guideline — leaving open rather than resolving. The itemised list is the surface in question, and its layout is being revised (see Blockers in the PR body), so building a ResponsiveComponent variant against the current chip layout would be thrown away. Will add one as part of that revision, or justify why this surface has no meaningful desktop/mobile divergence given it is a vertical row stack at every width.

AndrewAcomb:
Closing to re-implement as a stack for reviewability

AndrewAcomb:
Superseded by a stacked chain, split for reviewability:

- #1801 — extract `LegacyTaskBody` (pure move, no behaviour change)
- #1802 — model business tasks as a `task_type` discriminated union
- #1803 — `@ui/Chip` primitive
- #1804 — read and write paths
- #1805 — the answer UI

Same final tree as this PR's head `eccc184b8`, verified byte-identical to a squash of it. Each branch typechecks, lints and passes tests independently.

PR #1802, opened 2026-09-10
[LAY-2720] Split 2/5: model business tasks as a task_type discriminated union

AndrewAcomb:
## Scope

- Split `BusinessTaskSchema` into per-arm files under `businessTasks/`: the legacy human/upload shape, the counterparty ask shape, and a permissive catch-all.
- Add `isCounterpartyAskTask` / `isLegacyBusinessTask` / `isRenderableBusinessTask`, and drop unrecognised tasks in `getUserVisibleTasks`.
- Guard the task MSW handlers so they only patch legacy tasks.

The catch-all fixes a fragility that is live on `main`, not just a precaution for this feature. `AUTOMATED_RULE_SUGGESTION` and `AUTOMATED_TRANSACTION_REVIEW` serialize neither `question` nor `user_response_type`, and neither is `fromLlmTransactionCategorizer`, so `BookkeepingPeriod.deprecatedTasks` does not filter them out. One such task failed the previous single-struct decode and took the whole `/bookkeeping/periods` response down with it — blanking Tasks, the month selector and year status, including the legacy tasks that otherwise worked.

The legacy arm actively excludes the ask literal, so an ask can never decode as a free-response task and POST to `/user-response`, which the API rejects with a 400 for that type.

## Verification

- `npm run typecheck`, `npm run lint` clean.
- `npm test -- --run`: 353 tests pass.
- `businessTask.test.ts` covers both automated shapes decoding as unrenderable instead of throwing, and a malformed ask not passing as renderable.
- Arm shapes checked against `BusinessTasksRouter.kt` on `api` master.

Root: #1801
Base: #1801
Next: #1803

---

> [!NOTE]
> **Medium Risk**
> Changes how period task payloads decode and which tasks surface in the UI; incorrect union ordering or guards could hide tasks or mis-route mutations, though tests target the main failure modes.
> 
> **Overview**
> Replaces the single **legacy-only** `BusinessTaskSchema` with a **union** of counterparty-ask, legacy human/upload, and permissive unknown arms, plus shared base types and guards (`isLegacyBusinessTask`, `isCounterpartyAskTask`, `isRenderableBusinessTask`).
> 
> **User-visible tasks** now require a renderable legacy shape via `getUserVisibleTasks`, so automated and not-yet-built ask tasks no longer appear as broken TODO rows; the unknown arm also stops odd API payloads from failing the whole `/bookkeeping/periods` decode.
> 
> **MSW** task mutations route through `patchLegacyTaskInStore` so only legacy tasks are updated in the mock store (with fallback when the patch does not apply). Fixtures and `LegacyTaskBody` typing pick up `LegacyBusinessTask` / `task_type`. New Vitest coverage documents decode and visibility behavior for automated tasks and malformed asks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e5c83a25afd292c82a66cbbf33c78a5007ab496. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/components/features/bookkeeping/TasksListItem/TasksListItem.tsx
+++ b/src/components/features/bookkeeping/TasksListItem/TasksListItem.tsx
@@ -2,6 +2,7 @@ import { forwardRef, useCallback, useEffect, useState } from 'react'
 import classNames from 'classnames'
 
 import { LayerEventComponent, LayerEventType } from '@schemas/common/layerEvents'
+import { isLegacyBusinessTask } from '@schemas/features/bookkeeping/businessTask'
 import { isCompletedTask, type UserVisibleTask } from '@utils/features/bookkeeping/bookkeepingTasksFilters'
 import ChevronDownFill from '@icons/ChevronDownFill'
 import { useEmitLayerEvent } from '@hooks/utils/events/useEmitLayerEvent'
@@ -78,7 +79,9 @@ export const TasksListItem = forwardRef<HTMLDivElement, TasksListItemProps>((
           />
         </div>
         <div className={taskBodyClassName}>
-          <LegacyTaskBody task={task} onAnswered={onAnswered} />
+          {isLegacyBusinessTask(task)
+            ? <LegacyTaskBody task={task} onAnswered={onAnswered} />
+            : null}
         </div>
       </div>
     </div>
--- a/src/schemas/features/bookkeeping/businessTask.ts
+++ b/src/schemas/features/bookkeeping/businessTask.ts
@@ -1,65 +1,38 @@
-import { pipe, Schema } from 'effect'
+import { Schema } from 'effect'
 
-import { S3PresignedUrlSchema } from '@schemas/common/s3PresignedUrl'
-import { createTransformedEnumSchema } from '@schemas/common/utils'
+import { COUNTERPARTY_ASK_TASK_TYPE } from '@schemas/features/bookkeeping/businessTasks/baseBusinessTask'
+import {
+  type CounterpartyAskTask,
+  CounterpartyAskTaskSchema,
+} from '@schemas/features/bookkeeping/businessTasks/counterpartyAskTask'
+import {
+  type LegacyBusinessTask,
+  LegacyBusinessTaskSchema,
+} from '@schemas/features/bookkeeping/businessTasks/legacyBusinessTask'
+import { UnknownBusinessTaskSchema } from '@schemas/features/bookkeeping/businessTasks/unknownBusinessTask'
 
-export enum BusinessTaskStatus {
-  Todo = 'TODO',
-  UserMarkedCompleted = 'USER_MARKED_COMPLETED',
-  Completed = 'COMPLETED',
-  Archived = 'ARCHIVED',
-}
-
-export const BusinessTaskStatusSchema = Schema.Enums(BusinessTaskStatus)
-
-const TransformedBusinessTaskStatusSchema = createTransformedEnumSchema(
-  BusinessTaskStatusSchema,
-  BusinessTaskStatus,
-  BusinessTaskStatus.Todo,
+export const BusinessTaskSchema = Schema.Union(
+  CounterpartyAskTaskSchema,
+  LegacyBusinessTaskSchema,
+  UnknownBusinessTaskSchema,
 )
-
-export enum TaskUserResponseType {
-  FreeResponse = 'FREE_RESPONSE',
-  UploadDocument = 'UPLOAD_DOCUMENT',
-  Unknown = 'UNKNOWN',
-}
-
-export const TaskUserResponseTypeSchema = Schema.Enums(TaskUserResponseType)
-
-const TransformedTaskUserResponseTypeSchema = createTransformedEnumSchema(
-  TaskUserResponseTypeSchema,
-  TaskUserResponseType,
-  TaskUserResponseType.Unknown,
-)
-
-const TaskDocumentSchema = Schema.Struct({
-  fileName: pipe(
-    Schema.propertySignature(Schema.String),
-    Schema.fromKey('file_name'),
-  ),
-  presignedUrl: pipe(
-    Schema.propertySignature(S3PresignedUrlSchema),
-    Schema.fromKey('presigned_url'),
-  ),
-})
-
-// Every business task is treated as a human task; the automated variants are not
-// yet surfaced in the UI.
-export const BusinessTaskSchema = Schema.Struct({
-  id: Schema.UUID,
-  status: TransformedBusinessTaskStatusSchema,
-  title: Schema.String,
-  question: Schema.String,
-  userResponse: pipe(
-    Schema.propertySignature(Schema.NullishOr(Schema.String)),
-    Schema.fromKey('user_response'),
-  ),
-  userResponseType: pipe(
-    Schema.propertySignature(TransformedTaskUserResponseTypeSchema),
-    Schema.fromKey('user_response_type'),
-  ),
-  documents: Schema.NullishOr(Schema.Array(TaskDocumentSchema)),
-})
 
 export type BusinessTask = typeof BusinessTaskSchema.Type
 export type BusinessTaskEncoded = typeof BusinessTaskSchema.Encoded
+
+export const isCounterpartyAskTask = <T extends Pick<BusinessTask, 'taskType'>>(
+  task: T,
+): task is T & CounterpartyAskTask =>
+  task.taskType === COUNTERPARTY_ASK_TASK_TYPE && 'transactionResponses' in task
+
+// LegacyBusinessTask always carries user_response_type; UnknownBusinessTask (an
+// unrecognised task_type this union can't fully model) never does.
+export const isLegacyBusinessTask = <T extends Pick<BusinessTask, 'taskType'>>(
+  task: T,
+): task is T & LegacyBusinessTask =>
+  !isCounterpartyAskTask(task) && 'userResponseType' in task
+
+export const isRenderableBusinessTask = <T extends BusinessTask>(
+  task: T,
+): task is T & (CounterpartyAskTask | LegacyBusinessTask) =>
+  isCounterpartyAskTask(task) || isLegacyBusinessTask(task)
--- a/src/schemas/features/bookkeeping/businessTask.ts
+++ b/src/schemas/features/bookkeeping/businessTask.ts
@@ -20,19 +20,22 @@ export const BusinessTaskSchema = Schema.Union(
 export type BusinessTask = typeof BusinessTaskSchema.Type
 export type BusinessTaskEncoded = typeof BusinessTaskSchema.Encoded
 
-export const isCounterpartyAskTask = <T extends Pick<BusinessTask, 'taskType'>>(
+export const isCounterpartyAskTask = <T extends BusinessTask>(
   task: T,
 ): task is T & CounterpartyAskTask =>
   task.taskType === COUNTERPARTY_ASK_TASK_TYPE && 'transactionResponses' in task
 
 // LegacyBusinessTask always carries user_response_type; UnknownBusinessTask (an
 // unrecognised task_type this union can't fully model) never does.
-export const isLegacyBusinessTask = <T extends Pick<BusinessTask, 'taskType'>>(
+export const isLegacyBusinessTask = <T extends BusinessTask>(
   task: T,
 ): task is T & LegacyBusinessTask =>
   !isCounterpartyAskTask(task) && 'userResponseType' in task
 
+// CounterpartyAskTask has no renderable body yet (the Chip primitive lands in #1803, the
+// ask UI in #1805) — excluded here so it doesn't surface as a dead-end TODO row. Fold it
+// back in once that body ships.
 export const isRenderableBusinessTask = <T extends BusinessTask>(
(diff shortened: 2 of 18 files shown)


src/schemas/features/bookkeeping/businessTasks/unknownBusinessTask.ts
```diff
@@ -0,0 +1,16 @@
+import { pipe, Schema } from 'effect'
+
+import { TransformedBusinessTaskStatusSchema } from '@schemas/features/bookkeeping/businessTasks/baseBusinessTask'
+
+export const UnknownBusinessTaskSchema = Schema.Struct({
```
AndrewAcomb, line 5: Verified this against the actual backend serializers rather than the fixtures, and the fallback holds.

`UnknownBusinessTaskSchema` requires exactly `id`, `status`, `title`, `task_type`. The two shapes that previously had no viable union member both emit all four:

- `ApiBusinessTask.AutomatedTask` (`AUTOMATED_RULE_SUGGESTION`) — `id, task_run_id, title, status, period_id, effective_date, created_at, updated_at, completed_at, task_type, source_suggestion_id, suggestion`. No `question`, no `user_response_type`.
- `ApiBusinessTask.AutomatedTransactionReviewTask` — same minus the suggestion fields, plus `transactions`. Also no `question`.

So both now decode as `UnknownBusinessTask` instead of failing the whole periods payload, which was the concern. Keeping `question` required on `BaseBusinessTaskSchema` and out of the fallback is the right split — the fallback stays valid for task types that don't exist yet, and I do expect to add more.

Union ordering is also right: `Legacy` before `Unknown` means a `Human_Task` can't be captured by the looser member.

Two notes, neither blocking:

1. `BusinessTaskStatus.Completed = 'COMPLETED'` is dead against this API. The backend enum is `TODO`, `USER_MARKED_COMPLETED`, `ARCHIVED` only — `completed_at` is a timestamp, never a status value. Harmless since the transformed schema defaults unknown values, but it will never arrive.
2. Adding `'transactionResponses' in task` to `isCounterpartyAskTask` is a good call. A malformed ask that falls through to `Unknown` can't then be mistaken for a real one, which matters because the ask branch is the one that renders interactive controls.

src/schemas/features/bookkeeping/businessTask.ts
```diff
@@ -1,65 +1,38 @@
...
+  task.taskType === COUNTERPARTY_ASK_TASK_TYPE && 'transactionResponses' in task
+
+// LegacyBusinessTask always carries user_response_type; UnknownBusinessTask (an
+// unrecognised task_type this union can't fully model) never does.
+export const isLegacyBusinessTask = <T extends Pick<BusinessTask, 'taskType'>>(
+  task: T,
+): task is T & LegacyBusinessTask =>
+  !isCounterpartyAskTask(task) && 'userResponseType' in task
+
+export const isRenderableBusinessTask = <T extends BusinessTask>(
```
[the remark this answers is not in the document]
AndrewAcomb: Fixed in dd24119; filtered in getUserVisibleTasks.

src/utils/features/bookkeeping/bookkeepingTasksFilters.ts
```diff
@@ -1,4 +1,5 @@
-import { type BusinessTask, BusinessTaskStatus } from '@schemas/features/bookkeeping/businessTask'
+import { type BusinessTask } from '@schemas/features/bookkeeping/businessTask'
+import { BusinessTaskStatus } from '@schemas/features/bookkeeping/businessTasks/baseBusinessTask'
```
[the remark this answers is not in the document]
AndrewAcomb: Fixed in dd24119; filtered in getUserVisibleTasks.

src/schemas/features/bookkeeping/businessTasks/unknownBusinessTask.ts
AndrewAcomb: Thanks for verifying against the serializers.

AndrewAcomb:
augment review

AndrewAcomb:
cursor review

src/components/features/bookkeeping/TasksListItem/TasksListItem.tsx
```diff
@@ -78,7 +79,9 @@ export const TasksListItem = forwardRef<HTMLDivElement, TasksListItemProps>((
           />
         </div>
         <div className={taskBodyClassName}>
-          <LegacyTaskBody task={task} onAnswered={onAnswered} />
+          {isLegacyBusinessTask(task)
```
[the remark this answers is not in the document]
AndrewAcomb: Not reachable in this PR, and closed by the end of the stack. Asks cannot appear in the list here: the periods request still sends the API default legacy_tasks_only=true, which excludes every fromLlmTransactionCategorizer type, and the ask fixtures do not exist yet — the opt-in and the fixtures both land in #1804, and the body that renders them lands in #1805. So there is no merged state where an ask reaches this branch without a body. isRenderableBusinessTask admits the ask arm deliberately, because the schema knows the shape before the UI does.

sarahraines, line 82: Counterparty ask tasks now reach the list with no body.

`isRenderableBusinessTask` treats `CounterpartyAskTask` as renderable, so `getUserVisibleTasks` keeps asks, but nothing renders them yet — the ask body lands in a later PR in this stack (#1803 is the `Chip` primitive). As soon as the API emits `ASK_ABOUT_COUNTERPARTY_FOR_PERIOD`, a user gets a row that expands into an empty box: no question, no input, no way to answer. Worse, it stays `TODO` forever, so it counts in `getIncompleteTasks` — inflating the `TasksPending` count, the `TasksMonthSelector` unresolved badge and `useBookkeepingYearsStatus`, and keeping the period permanently "awaiting customer".

Until the ask UI exists, consider having `isRenderableBusinessTask` accept only legacy tasks (asks then decode fine — which is the actual fix for the periods-response blowup — but stay out of the list), and flip it on in the PR that adds the body.

src/schemas/features/bookkeeping/businessTasks/counterpartyAskTask.ts
```diff
@@ -0,0 +1,91 @@
...
+export type CounterpartyAskTransactionResponse = typeof CounterpartyAskTransactionResponseSchema.Type
+
+export const CounterpartyAskTaskSchema = Schema.extend(
+  BaseBusinessTaskSchema,
+  Schema.Struct({
+    taskType: pipe(
+      Schema.propertySignature(Schema.Literal(COUNTERPARTY_ASK_TASK_TYPE)),
+      Schema.fromKey('task_type'),
+    ),
+    counterparty: Schema.NullishOr(BankTransactionCounterpartySchema),
```
sarahraines, line 47: The nested ask shapes are stricter than the arm they gate, and a mismatch is now silent.

Because the union falls through to `UnknownBusinessTaskSchema`, any decode failure inside the ask arm no longer surfaces as an error — the task quietly becomes unrenderable and `getUserVisibleTasks` drops it. Several nested requirements can trigger that on a perfectly legitimate ask:

- `BankTransactionCounterpartySchema` requires `mccs: Schema.Array(Schema.String)` (required key, not nullable) — a counterparty serialized without `mccs`, or with `mccs: null`, drops the whole task.
- `MinimalBankTransactionSchema.direction` only accepts `CREDIT|MONEY_IN|DEBIT|MONEY_OUT`, and `date` must parse as a `Date`; one unexpected value in `transactions` drops the task.
- `suggestions[].account_identifier` must be exactly `StableName` or `AccountId`; a third identifier type drops the task.

Before this PR such a payload failed loudly; now the ask just never appears and there is no log to explain it. Worth loosening the nested nullable fields (e.g. `mccs` as `optionalWith(..., {default: () => [], nullable: true})`) and/or emitting a dev-mode warning when a task decodes into the `Unknown` arm, so silent drops are observable.

src/schemas/features/bookkeeping/businessTasks/unknownBusinessTask.ts
sarahraines, line 5: The catch-all is not catch-all enough to guarantee the periods response decodes.

`UnknownBusinessTaskSchema` requires `task_type` (required, non-nullable `String`), `title`, a UUID `id`, and a non-empty-trimmed `status`. A task missing any of those still fails all three union arms, and because `tasks: Schema.Array(BusinessTaskSchema)` sits inside `BookkeepingPeriodsSchema`, that one task again takes down the entire `/bookkeeping/periods` response — exactly the failure this PR sets out to eliminate.

Note the two arms disagree about the same field: `LegacyBusinessTaskSchema` deliberately makes `task_type` optional with a `null` default (i.e. it may be absent), while this arm makes it required. So a task with no `task_type` that also lacks `question`/`user_response_type` matches nothing.

Suggest making the catch-all tolerant of everything except `id`:

```suggestion
export const UnknownBusinessTaskSchema = Schema.Struct({
  id: Schema.UUID,
  status: TransformedBusinessTaskStatusSchema,
  title: Schema.optionalWith(Schema.String, { default: () => "", nullable: true }),
  taskType: pipe(
    Schema.optionalWith(Schema.NullishOr(Schema.String), { default: () => null }),
    Schema.fromKey("task_type"),
  ),
})
```

src/schemas/features/bookkeeping/businessTask.ts
```diff
@@ -1,65 +1,38 @@
...
 export type BusinessTaskEncoded = typeof BusinessTaskSchema.Encoded
+
+export const isCounterpartyAskTask = <T extends Pick<BusinessTask, 'taskType'>>(
+  task: T,
+): task is T & CounterpartyAskTask =>
+  task.taskType === COUNTERPARTY_ASK_TASK_TYPE && 'transactionResponses' in task
+
+// LegacyBusinessTask always carries user_response_type; UnknownBusinessTask (an
+// unrecognised task_type this union can't fully model) never does.
+export const isLegacyBusinessTask = <T extends Pick<BusinessTask, 'taskType'>>(
```
sarahraines: These guards are sound only for fully decoded tasks, but their signature advertises much less.

`isLegacyBusinessTask` accepts any `T extends Pick<BusinessTask, "taskType">` and decides via `!isCounterpartyAskTask(task) && "userResponseType" in task`, while `isCounterpartyAskTask` additionally requires `"transactionResponses" in task`. For an object that carries `taskType` and `userResponseType` but not `transactionResponses` — a projection, a hand-built fixture, a partially-constructed task — an `ASK_ABOUT_COUNTERPARTY_FOR_PERIOD` task classifies as *legacy*: `isCounterpartyAskTask` returns `false` for lack of `transactionResponses`, so `isLegacyBusinessTask` returns `true`. `TasksListItem` would then mount `LegacyTaskBody` for an ask and POST to `/user-response`, which the PR description says the API rejects with a 400 for that task type.

Today no call site does this, but nothing in the types stops one. Constraining both guards to `T extends BusinessTask` (as `isRenderableBusinessTask` already does) makes the presence checks match what the signature promises.

src/schemas/features/bookkeeping/businessTask.ts
```diff
@@ -20,19 +20,22 @@ export const BusinessTaskSchema = Schema.Union(
...
 // unrecognised task_type this union can't fully model) never does.
-export const isLegacyBusinessTask = <T extends Pick<BusinessTask, 'taskType'>>(
+export const isLegacyBusinessTask = <T extends BusinessTask>(
   task: T,
 ): task is T & LegacyBusinessTask =>
   !isCounterpartyAskTask(task) && 'userResponseType' in task
 
+// CounterpartyAskTask has no renderable body yet (the Chip primitive lands in #1803, the
+// ask UI in #1805) — excluded here so it doesn't surface as a dead-end TODO row. Fold it
+// back in once that body ships.
```
sarahraines: Remove this useless comment.

sarahraines (approved):

src/components/features/bookkeeping/TasksListItem/TasksListItem.tsx
AndrewAcomb: Agreed — this landed on main: isRenderableBusinessTask is legacy-only, and #1805 folds asks back in now the body ships.

src/schemas/features/bookkeeping/businessTasks/unknownBusinessTask.ts
AndrewAcomb: Fixed on main — the catch-all now only requires id.

src/schemas/features/bookkeeping/businessTask.ts
AndrewAcomb: Fixed on main — both guards are T extends BusinessTask via Schema.is.

src/schemas/features/bookkeeping/businessTasks/counterpartyAskTask.ts
AndrewAcomb: Fixed in c5c55f39d on #1804: mccs is now optional/nullable with a [] default, and getUserVisibleTasks warns once per task id when one decodes into the Unknown arm, so silent drops are observable. Left direction/account_identifier strict — narrowing those is a real decode error worth seeing.

PR #1811, opened 2026-09-17
[LAY-2729] Reconnection workflow: domain utils, schema, and shared hooks (1/5)

darrensapalo:
## Scope
- Adds bank-account reconnection eligibility logic (`getBankAccountNeedingReconnection`, `getBankAccountRefreshConnections`, staleness helpers) in `bankAccount.ts`
- Adds `lastSyncedAt`/`updateType` schema fields these read, plus matching fixtures
- Adds two shared hooks used by the UI layers stacked on top: `usePeriodicNow`, `useIsMobileContainer`
- Adds idempotent-provider detection to `LinkedAccountsContext` (`useHasLinkedAccountsProvider`)

## Verification
- `npm run typecheck`, `npm run lint`
- `bankAccount.test.ts` covers the new eligibility/selector functions

Root: this
Base: main
Next: #1812

---

> [!NOTE]
> **Medium Risk**
> Touches external-account schema and linked-account reconnect routing; behavior is mostly new selectors and tests, but incorrect staleness or grouping logic could show or hide refresh prompts wrongly.
> 
> **Overview**
> Introduces **domain support for bank reconnection/refresh** ahead of UI work: external account connections now carry `lastSyncedAt` and `updateType` (`BACKGROUND`, `USER_PRESENT_REQUIRED`, `UNKNOWN`), and `reconnectWithNewCredentials` is nullable on the schema.
> 
> Adds **`refresh.ts`** selectors that treat connections as needing user-present refresh when `updateType` is `USER_PRESENT_REQUIRED`, the link has a `connectionExternalId`, and sync is missing or older than 24 hours—grouping by connection for multi-account alerts and exposing single-account `getBankAccountNeedingReconnection`. **Syncing helpers** move from `bankAccount.ts` into **`sync.ts`**; **`getBankAccountLabelParts`** supports display metadata in refresh payloads.
> 
> Adds **`useReconnectConnection`** to call `addConnection` vs `repairConnection` from `LinkedAccountsContext`, **`useHasLinkedAccountsProvider`** to detect a real provider, fixture helpers (`markAccountReadyForRefresh`), MSW clearing `updateType` after connection status updates, and **`refresh.test.ts`** coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9b59c12a6c0bbc0c5cac82e7c061650efea81a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/hooks/utils/size/useIsMobileContainer.ts
```diff
@@ -0,0 +1,29 @@
...
+      if (!entry) return
+
+      const width = entry.borderBoxSize[0]?.inlineSize ?? entry.contentRect.width
+      setIsMobile(width <= BREAKPOINTS.MOBILE)
+    })
+    observer.observe(element)
+    observerRef.current = observer
+  }, [])
+
+  return { isMobile, containerRef }
```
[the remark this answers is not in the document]
darrensapalo: Fixed in [6ca0b2535](https://github.com/Layer-Fi/layer-react/commit/6ca0b25355d670b2af4fdbca5d7a30bd21bd9f96) — wraps the return value in `useMemo`.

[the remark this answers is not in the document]
darrensapalo: Correction: the fix commit is [6ca0b2535](https://github.com/Layer-Fi/layer-react/commit/6ca0b25359487c178981f0d4e01d31f58ad2cde7) (the link in my previous reply had a typo'd hash) — wraps the return value in `useMemo`.

src/hooks/utils/size/useIsMobileContainer.ts
```diff
@@ -0,0 +1,29 @@
+import { useCallback, useMemo, useRef, useState } from 'react'
+
+import { BREAKPOINTS } from '@utils/shared/size/screenSizeBreakpoints'
+
+export function useIsMobileContainer<T extends HTMLElement>() {
```
doneel: I'm biased towards not introducing `useIsMobileContainer` here and in the follow-on PRs, use `useElementViewSize and view === 'mobile'` (matching Reports), or `useElementSize`.

It seems unlikely to me an _almost the same_ hook is a good idea to add here.

doneel (approved):

src/hooks/utils/size/useIsMobileContainer.ts
darrensapalo: Removed in 13e33aa

src/hooks/utils/dates/usePeriodicNow.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from 'react'
+
+const DEFAULT_INTERVAL_MS = 5 * 60 * 1000
+
+export function usePeriodicNow(intervalMs: number = DEFAULT_INTERVAL_MS): Date {
```
sarahraines: This is not necessary - we already poll every 10 minutes due to our default SWR config.

src/utils/features/bankAccounts/bankAccount.ts
```diff
@@ -25,6 +35,122 @@ export function isBankAccountSyncing(bankAccount: BankAccount): boolean {
...
+  now = new Date(),
+): BankAccountNeedingReconnection | null {
+  for (const bankAccount of bankAccounts ?? []) {
+    const refreshInfo = getBankAccountRefreshConnectionInfo(bankAccount, now)
+    if (refreshInfo) {
+      return {
+        accountLabel: formatBankAccountWithMask(bankAccount),
+        lastSyncedAt: refreshInfo.lastSyncedAt ?? null,
+        source: refreshInfo.source,
+        connectionExternalId: refreshInfo.connectionExternalId ?? null,
```
sarahraines: **Medium — this can render a dead "Reconnect" button.**

`connectionExternalId` is typed `string | null` here, but the sibling `getBankAccountRefreshConnections` explicitly skips external accounts with no id (line 63), and `LinkedAccountGridItem.tsx:106` guards with `if (!repairInfo.connectionExternalId) return` — so null ids do occur in practice.

An account with `USER_PRESENT_REQUIRED` and a null `connectionExternalId` therefore reaches the single-account surface and renders a CTA that can never call `repairConnection`.

Folding `isRepairable` into the shared `isDueForRefresh` predicate (see the review summary) removes the `?? null` and narrows the field to `string`, so this can't be reached.

src/utils/features/bankAccounts/bankAccount.ts
```diff
@@ -25,6 +35,122 @@ export function isBankAccountSyncing(bankAccount: BankAccount): boolean {
   return bankAccount.externalAccounts.some(ea => ea.isSyncing)
 }
 
+export function isBankAccountReadyForRefresh(bankAccount: BankAccount, now = new Date()): boolean {
```
sarahraines: **Medium — the count and the list disagree on eligibility.**

This and `getBankAccountsReadyForRefresh` go straight to `isExternalAccountDueForRefresh` with no `connectionExternalId` check, while `getBankAccountRefreshConnections` drops id-less accounts at line 63.

Gate a banner on this count and render the grouped list beneath it and you get "1 account needs reconnection" above zero rows. Both entry points should share one predicate.

src/utils/features/bankAccounts/bankAccount.ts
```diff
@@ -25,6 +35,122 @@ export function isBankAccountSyncing(bankAccount: BankAccount): boolean {
...
+    connectionExternalId: string
+    source: ExternalAccountConnection['externalAccountSource']
+    reconnectWithNewCredentials: boolean
+    institutionNames: Set<string>
+    accountNames: Set<string>
+  }>()
+
+  for (const bankAccount of bankAccounts ?? []) {
+    for (const externalAccount of bankAccount.externalAccounts) {
+      if (!externalAccount.connectionExternalId || !isExternalAccountDueForRefresh(externalAccount, now)) continue
```
sarahraines: This is the only place the `connectionExternalId` requirement is applied. Worth moving into the shared `isDueForRefresh` predicate so `isBankAccountReadyForRefresh` (line 38) and `getBankAccountNeedingReconnection` (line 125) can't disagree with it.

src/utils/features/bankAccounts/bankAccount.ts
```diff
@@ -25,6 +35,122 @@ export function isBankAccountSyncing(bankAccount: BankAccount): boolean {
...
+      const key = JSON.stringify([externalAccount.externalAccountSource, externalAccount.connectionExternalId])
+      const connection = connections.get(key) ?? {
+        connectionExternalId: externalAccount.connectionExternalId,
+        source: externalAccount.externalAccountSource,
+        reconnectWithNewCredentials: false,
+        institutionNames: new Set<string>(),
+        accountNames: new Set<string>(),
+      }
+
+      const institutionName = externalAccount.institution?.name ?? bankAccount.institution?.name
```
sarahraines: **Institution precedence is inverted relative to the existing helper.**

`getBankAccountInstitution` (line 29) prefers `bankAccount.institution` and falls back to the external account; this line does the opposite. When the two differ — a mirrored account whose bank-account institution was renamed — the reconnection alert names a different bank than the account card next to it.

```suggestion
      const institutionName = getBankAccountInstitution(bankAccount)?.name ?? externalAccount.institution?.name
```

src/utils/features/bankAccounts/bankAccount.ts
```diff
@@ -25,6 +35,122 @@ export function isBankAccountSyncing(bankAccount: BankAccount): boolean {
...
+      if (institutionName) connection.institutionNames.add(institutionName)
+      connection.accountNames.add(formatBankAccountWithMask(bankAccount))
```
sarahraines: **Dedupes on the rendered label, which under-reports the account count.**

`accountNames` is a `Set<string>` of the formatted string, so two distinct sub-accounts that format identically (same name, same mask — e.g. two "Checking (0000)" entries) collapse into one and the alert reports fewer affected accounts than there are.

Group by `bankAccount.id` and format afterwards, if dedupe is wanted at all.

src/utils/features/bankAccounts/bankAccount.ts
```diff
@@ -25,6 +35,122 @@ export function isBankAccountSyncing(bankAccount: BankAccount): boolean {
...
+        connectionExternalId: refreshInfo.connectionExternalId ?? null,
+        reconnectWithNewCredentials: refreshInfo.reconnectWithNewCredentials,
+      }
+    }
+  }
+
+  return null
+}
+
+function isExternalAccountReadyForRefresh(externalAccount: ExternalAccountConnection): boolean {
```
sarahraines: **Name promises more than it checks.**

Public `isBankAccountReadyForRefresh` (line 38) includes the 24-hour staleness window; this one is *only* the `UserPresentRequired` check — staleness lives in `isExternalAccountRefreshStale`. A future caller reaching for the external-level "ReadyForRefresh" would silently bypass the staleness guard.

Suggest naming it for what it does, e.g. `isExternalAccountUserPresentRequired`.

src/utils/features/bankAccounts/bankAccount.ts
```diff
@@ -25,6 +35,122 @@ export function isBankAccountSyncing(bankAccount: BankAccount): boolean {
...
+function isExternalAccountRefreshStale(externalAccount: ExternalAccountConnection, now: Date): boolean {
+  return externalAccount.lastSyncedAt == null
+    || now.getTime() - externalAccount.lastSyncedAt.getTime() > REFRESH_ALERT_MINIMUM_SYNC_AGE_MS
+}
+
+function isExternalAccountDueForRefresh(externalAccount: ExternalAccountConnection, now: Date): boolean {
+  return isExternalAccountReadyForRefresh(externalAccount) && isExternalAccountRefreshStale(externalAccount, now)
+}
+
+export function formatBankAccountWithMask(bankAccount: BankAccount): string {
```
sarahraines: **Untranslated user-visible string.**

This builds `${accountName} (${mask})` by concatenation with no `t()`, which the strings non-negotiable in AGENTS.md covers — and it's the value returned as `accountLabel`, so it renders directly in parts 2–5.

Since a pure util can't hold the `t` instance, the cleanest fix is probably to return the parts and format at the render site. Cheaper to settle now than after three call sites depend on the raw shape.

src/utils/features/bankAccounts/refresh.test.ts
```diff
@@ -0,0 +1,194 @@
...
+  getBankAccountRefreshConnections,
+  getBankAccountsReadyForRefresh,
+  isBankAccountReadyForRefresh,
+} from '@utils/features/bankAccounts/bankAccount'
+
+import {
+  makeBankAccount,
+  makeBankAccountWithMirroredExternalAccount,
+} from '@fixtures/bankAccounts/mocks'
+import { MS_PER_HOUR, NOW } from '@testUtils/dates/fixedDates'
```
sarahraines, line 16: `NOW` is imported here, but `setupFakeSystemTime(NOW)` (`@testUtils/dates/fakeSystemTime`) is never called — so the tests that rely on the `now = new Date()` defaults run against the real clock while the later ones use the 2026-06-15 fixture.

They pass today only because those fixtures leave `lastSyncedAt` null. Adding a `lastSyncedAt` to one would make the suite clock-dependent and flaky. Worth pinning the clock now.

sarahraines:
Deep review of the domain layer. The schema and fixture work looks consistent — every `ExternalAccountConnection` literal in the repo was updated, and loosening `reconnectWithNewCredentials` to `NullishOr` breaks no caller (both consumers in `LinkedAccountGridItem.tsx` use truthiness).

The substantive feedback is one theme rather than eight separate nits: **three functions each walk `bankAccounts → externalAccounts` and each re-decides eligibility inline**, and they don't agree with each other. Rather than patch each site, I'd suggest a small set of subhelpers so the divergence becomes unrepresentable:

```ts
type RepairableExternalAccount = ExternalAccountConnection & { connectionExternalId: string }

const isUserPresentRequired = (ea: ExternalAccountConnection) =>
  ea.updateType === ExternalAccountUpdateType.UserPresentRequired

const isSyncStale = (ea: ExternalAccountConnection, now: Date) =>
  ea.lastSyncedAt == null || now.getTime() - ea.lastSyncedAt.getTime() > REFRESH_ALERT_MINIMUM_SYNC_AGE_MS

const isRepairable = (ea: ExternalAccountConnection): ea is RepairableExternalAccount =>
  ea.connectionExternalId != null

const isDueForRefresh = (ea: ExternalAccountConnection, now: Date): ea is RepairableExternalAccount =>
  isUserPresentRequired(ea) && isSyncStale(ea, now) && isRepairable(ea)
```

plus one shared traversal:

```ts
type RefreshableAccount = { bankAccount: BankAccount, externalAccount: RepairableExternalAccount }

function getRefreshableAccounts(
  bankAccounts: ReadonlyArray<BankAccount> | undefined,
  now: Date,
): RefreshableAccount[] {
  return (bankAccounts ?? []).flatMap(bankAccount =>
    bankAccount.externalAccounts
      .filter(ea => isDueForRefresh(ea, now))
      .map(externalAccount => ({ bankAccount, externalAccount })),
  )
}
```

Every public selector then reduces to a one-liner over it: `isBankAccountReadyForRefresh` is `.some`, `getBankAccountsReadyForRefresh` a unique-by-id map, `getBankAccountRefreshConnections` a group-by on `[source, connectionExternalId]`, `getBankAccountNeedingReconnection` an `[0]`. The payoff is on that last one — because the element type carries `connectionExternalId: string`, the `?? null` disappears and the field narrows to `string`, so the dead-CTA bug below stops being expressible rather than being guarded against.

One smaller helper alongside those, for the institution mismatch:

```ts
const getRefreshInstitutionName = ({ bankAccount, externalAccount }: RefreshableAccount) =>
  (getBankAccountInstitution(bankAccount)?.name ?? externalAccount.institution?.name) ?? null
```

routing through the existing `getBankAccountInstitution` rather than re-deriving the precedence.

Per-site inline comments follow in a separate review. Nothing here is a blocker for a domain-only PR, but the two mediums will surface as real UI bugs once parts 2–5 consume these.

